PR #5, opened 2022-11-25
Update router to new path patterns

kamorel:
1. Updates router to use the decided on path patterns. 
2. Updated base .vue file names to match these patterns.
3. Split BucketsList away HomeView so it's no longer the first thing a user sees.


frontend/src/components/bucket/BucketsTable.vue
```diff
@@ -3,7 +3,7 @@ import { storeToRefs } from 'pinia';
 import DataTable from 'primevue/datatable';
 import Column from 'primevue/column';
 import Button from 'primevue/button';
-
+import { RouteNames } from '@/utils/constants';
 import { useBucketStore } from '@/store/bucketStore';
```
jujaga, line 7: Just as an optional style note - when we're importing from our own code, we've historically alphabetized based on the from path; so in this case this line should be below the useBucketStore as store comes before utils. Don't need to follow this pattern everywhere but I figure it's not a bad thing to continue doing.
For the primevue imports, we should alphabetize this if possible too.

frontend/src/components/bucket/BucketList.vue
```diff
@@ -26,8 +18,12 @@ const closeInfo = () => {
...
 };
 
-onMounted(() => {
-  loadBuckets();
+onMounted(async () => {
+  try {
+    await bucketStore.load();
+  } catch (error: any) {
+    toast.add({ severity: 'error', summary: 'Unable to load buckets.', detail: error, life: 5000 });
+  }
```
jujaga, line 26: Ref toaster consideration.

kamorel: Added useToast composable

jujaga (approved):
🍞

frontend/src/composables/useToaster.ts
```diff
@@ -0,0 +1,13 @@
...
+export async function useToaster(callback: Function, options: any = {}) {
+  const toast = useToast();
+  try {
+    await callback();
+  } catch (error: any) {
+    const { severity = 'error', summary = 'Unable to load data.', detail = error, life = 5000 } = options;
+    toast.add({ severity: severity, summary: summary, detail: detail, life: life });
+  }
+
+  return;
```
loneil, line 12: Just noting for future use we could take in an array of callbacks and await each, and also expose a `loading` from this composable so the parent using it would know when the function(s) have completed. Can think about putting that stuff in later.

loneil (approved):

TimCsaky (approved):

PR #6, opened 2022-11-29
Make refresh token work

loneil:
Computed properties tied to the Keycloak object seem to not update no matter what I do... so the token getter is still returning the original token even though the refresh loop is going fine and getting a new token into the Keycloak object.
If we just expose the Keycloak object itself as a getter and then reference its properties in components that works fine though.

So for now we should just live with this to get the refresh token and other Keycloak property access working as expected.
Or maybe it's fine to have the one object exposed, it's TypeScript now and it's typed so it's not like you can reference things that aren't there or wouldn't work...

Can come back to later as well and try some more.

### DIFF
--- a/frontend/src/components/Login.vue
+++ b/frontend/src/components/Login.vue
@@ -5,18 +5,18 @@ import Button from 'primevue/button';
 import { useAuthStore } from '@/store/authStore';
 
 const authStore = useAuthStore();
-const { ready, getToken } = storeToRefs(useAuthStore());
+const { ready, getKeycloak } = storeToRefs(useAuthStore());
 
 const login = () => {
   authStore.login();
 };
 
-const logout = async () => {
-  const response = await authStore.logout();
+const logout = () => {
+  authStore.logout();
 };
 </script>
 
 <template>
-  <Button v-if="!getToken" primary @click="login" :loading="!ready">Log in</Button>
+  <Button v-if="!getKeycloak.token" primary @click="login" :loading="!ready">Log in</Button>
   <Button v-else primary @click="logout">Log out</Button>
 </template>
--- a/frontend/src/components/Login.vue
+++ b/frontend/src/components/Login.vue
@@ -4,16 +4,8 @@ import { storeToRefs } from 'pinia';
 import Button from 'primevue/button';
 import { useAuthStore } from '@/store/authStore';
 
-const authStore = useAuthStore();
+const { login, logout } = useAuthStore();
 const { ready, getKeycloak } = storeToRefs(useAuthStore());
-
-const login = () => {
-  authStore.login();
-};
-
-const logout = () => {
-  authStore.logout();
-};
 </script>
 
 <template>
--- a/frontend/src/services/interceptors.ts
+++ b/frontend/src/services/interceptors.ts
@@ -21,7 +21,7 @@ export function comsAxios(timeout = 10000) {
   instance.interceptors.request.use(
     (cfg: any) => {
       if (authStore.ready && authStore.getAuthenticated) {
-        cfg.headers.Authorization = `Bearer ${authStore.getToken}`;
+        cfg.headers.Authorization = `Bearer ${authStore.getKeycloak.token}`;
       }
       return Promise.resolve(cfg);
     },
--- a/frontend/src/services/interceptors.ts
+++ b/frontend/src/services/interceptors.ts
@@ -20,7 +20,7 @@ export function comsAxios(timeout = 10000) {
 
   instance.interceptors.request.use(
     (cfg: any) => {
-      if (authStore.ready && authStore.getAuthenticated) {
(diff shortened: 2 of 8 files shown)


frontend/src/store/authStore.ts
```diff
@@ -43,36 +44,46 @@ export const useAuthStore = defineStore('auth', () => {
...
 
-        //Token Refresh
+        // Token Refresh
+        // Check token validity every 10s and, if necessary, update the token.
+        // Refresh token if it's valid for less then 70 seconds
         setInterval(() => {
-          $keycloak.value
-            .updateToken(70)
+          $keycloak.value;
+          kc.updateToken(290) // If the token expires within 70 seconds from now get a refreshed
```
kamorel: Wrong min validity?

frontend/src/services/interceptors.ts
```diff
@@ -21,7 +21,7 @@ export function comsAxios(timeout = 10000) {
   instance.interceptors.request.use(
     (cfg: any) => {
       if (authStore.ready && authStore.getAuthenticated) {
```
kamorel, line 23: Should getAuthenticated be authStore.getKeycloak.authenticated instead?

frontend/src/store/authStore.ts
```diff
@@ -13,6 +13,7 @@ export const useAuthStore = defineStore('auth', () => {
   const ready = ref(false);
 
   // Getters
+  const getKeycloak = computed(() => $keycloak.value);
   const getAuthenticated = computed((): boolean => (ready.value ? !!$keycloak.value.authenticated : false));
   const getLoginUrl = computed(() => (ready.value ? $keycloak.value.createLoginUrl() : ''));
   const getToken = computed(() => (ready.value ? $keycloak.value.token : ''));
```
kamorel, line 19: If everything is going through getKeycloak now, maybe we can remove these other 3 getters?

frontend/src/services/interceptors.ts
loneil: Yeah good catch (this would work anyways because the authenticated stays set until you log out, whereas refresh token changes, but should update to be consistent)

frontend/src/components/Login.vue
```diff
@@ -5,18 +5,18 @@ import Button from 'primevue/button';
...
 
 const login = () => {
   authStore.login();
 };
 
-const logout = async () => {
-  const response = await authStore.logout();
+const logout = () => {
+  authStore.logout();
 };
```
jujaga, line 16: Do these need to be encapsulated as arrow functions? They may be invokable directly by just doing...
```js
const login = authStore.login();
const logout = authStore.logout();
```
That way it removes the need to have an intermediary anonymous function just to invoke the same thing. Don't know if that collides with Vue3 best practices or not.

frontend/src/store/authStore.ts
```diff
@@ -43,36 +44,45 @@ export const useAuthStore = defineStore('auth', () => {
...
         $keycloak.value = kc;
 
-        //Token Refresh
+        // Token Refresh
+        // Check token validity every 10s and, if necessary, update the token.
+        // Refresh token if it's valid for less then 70 seconds
         setInterval(() => {
-          $keycloak.value
-            .updateToken(70)
+          kc.updateToken(70) // If the token expires within 70 seconds from now get a refreshed
```
jujaga, line 67: 1) I can't remember, but we should make sure this setInterval is saved into auth state somehow. IIRC, setInterval returns an integer which is an identifier for the periodic function. We'll want to make sure that even though we don't use `clearInterval(identityId)`, we can choose to do so in the future if we need to by having it be assigned somewhere pragmatic in the auth store. Perhaps we should have it be called like `const refreshJob = setInterval((...` or something?

2) Magic numbers like that 70 seconds should be constant defined in a constants.js file.

jujaga (commented):
$keycloak -> _keycloak rename is appropriate as the dollar sign implies vue global, which is not something we're doing here. Underscore is good since it'll imply private attribute/component, which is more in line with what it's acting as now.

kamorel (approved):

frontend/src/components/Login.vue
loneil: Yeah that's the good Pinia way of doing it anways, there was some unneeded await there so we can simplify now 👍

frontend/src/store/authStore.ts
loneil: I saved the number in state (needed to change to window.setInterval or else it returns some other type that TS is wierd about https://stackoverflow.com/questions/53189729/typescript-error-setinterval-type-timer-is-not-assignable-to-type-number)
Let me know it that's what you had in mind?

TimCsaky (approved):

dc-bcgov (approved):

jujaga (approved):

PR #9, opened 2022-12-07
Object list view & global updates

kamorel:
View changes to further match design spec
- Updated table columns
- Implements info action icon and side panel

Also includes
- scss updates to move many duplicate scoped items to a global level
- Implements the PrimeVue ConfirmationDialog component at a global level


frontend/src/components/bucket/BucketList.vue
```diff
@@ -17,9 +17,7 @@ const closeInfo = () => {
   displayInfo.value = null;
 };
 
-onMounted(() => {
-  useToaster(bucketStore.load, { summary: 'Unable to load buckets.' });
-});
+await useToaster(bucketStore.load, { summary: 'Unable to load buckets.' });
```
loneil, line 20: Think this was just an experiment while screen sharing? Could keep back in onMounted for now

frontend/src/components/bucket/BucketsTable.vue
```diff
@@ -51,7 +51,7 @@ const showInfo = async (id: number) => {
         </div>
       </template>
     </Column>
-    <Column header="Actions" headerStyle="width: 12%" headerClass="header-right" bodyClass="content-right">
+    <Column header="Actions" headerStyle="width: 12%" headerClass="header-right" bodyClass="content-right actions-buttons">
```
jujaga, line 54: Any way to make this column always have a fixed size instead of relative? On relatively smaller viewports the buttons begin flowing around oddly.

frontend/src/components/object/ObjectTable.vue
```diff
@@ -33,60 +44,54 @@ const { loading, objectList } = storeToRefs(useObjectStore());
...
+        </div>
+      </template>
+    </Column>
+    <Column header="Updated date" :hidden="displayInfo ? true : false">
       <template #body="{ data }">
         {{ formatDateLong(data.updatedAt) }}
       </template>
     </Column>
+    <Column header="Actions" headerStyle="width: 15%" headerClass="header-right" bodyClass="content-right actions-buttons">
+      <template #body="{ data }">
```
jujaga, line 81: Any way to make this column always have a fixed size instead of relative? On relatively smaller viewports the buttons begin flowing around oddly.

frontend/src/components/bucket/BucketsTable.vue
kamorel: Fix added in PR-8

frontend/src/components/object/ObjectTable.vue
kamorel: Similar to the fix in PR-8. Forced column to be 200px.

jujaga (approved):

TimCsaky (approved):

PR #10, opened 2022-12-12
Download file buttons

loneil:
# Description

Implement Download file buttons. As discussed can mostly treat this as single file functionality, though if you allow popups it will do multiple selected
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2917

Exact download mechanism will be affected by decisions above so just doing tab opens of returned S3 url for this implementation.

Stub out delete/delete multi functionality. Will add in API calls in a next PR if the patterns of the buttons in this PR look good to us. I'll try and add in some unit tests in that PR too if we like this pattern for the buttons.

## Types of changes

## Checklist

## Further comments

image

image


frontend/src/components/object/ObjectTable.vue
```diff
@@ -79,18 +79,14 @@ const showInfo = async (id: string) => {
     </Column>
     <Column header="Actions" headerStyle="width: 200px" headerClass="header-right" bodyClass="content-right actions-buttons">
       <template #body="{ data }">
-        <Button class="p-button-lg p-button-text">
-          <font-awesome-icon icon="fa-solid fa-download" />
-        </Button>
+        <DownloadObjectButton :mode="ButtonMode.ICON" :ids="['1']" />
```
TimCsaky: is :ids="['1']" a placeholder?
are you going to do something like :ids="[data.id]" and then if multiple are selected push them all to a single array?

frontend/src/components/object/DeleteObjectButton.vue
```diff
@@ -0,0 +1,50 @@
...
+// Deletion
+const confirmDelete = () => {
+  confirm.require({
+    message: `Please confirm that you want to delete the selected ${props.ids.length > 1 ? 'files' : 'file'}`,
+    header: `Delete ${props.ids.length > 1 ? 'items' : 'item'}`,
+    acceptLabel: 'Confirm',
+    rejectLabel: 'Cancel',
+    accept: () => {
+      // TODO: Delete items
+      alert(`Not Implemented Yet (IDs ${props.ids})`);
```
TimCsaky: i guess you might have to update the store to remove the objects after their deleted.
by the way, by default calling the COMS DELETE /object/:objectId does a 'soft delete'..
i dont think we need to let users recover soft-deletes through the UI.. Maybe for now we only do hard-deletes which would be done by calling DELETE /object/:objectId?versionId=<latest version>
there's more info here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/DeletingObjectVersions.html

TimCsaky (approved):

frontend/src/components/object/DeleteObjectButton.vue
loneil: Yeah delete will likely wait until all are done and then re-fetch the table. Implementing next and will see what to do.

frontend/src/components/object/ObjectList.vue
```diff
@@ -28,6 +33,12 @@ const closeInfo = () => {
   displayInfo.value = null;
 };
 
+// Download
+const mutiSelectedObjectIds = computed(() => {
```
TimCsaky, line 37: not sure if you missed an 'L' in this

frontend/src/components/object/DownloadObjectButton.vue
```diff
@@ -0,0 +1,53 @@
...
+  },
+});
+
+// Download the object(s)
+const displayNoFileDlg = ref(false);
+const download = () => {
+  if (props.ids.length) {
+    // For now we are looping the supplied IDs (if multiple selected) until there is a batching feature
+    props.ids.forEach((i) => {
+      objectStore.getObject(i, undefined, true);
```
TimCsaky, line 33: i'm guessing this shouldnt be awaiting calls to an async getObject
not sure if or where you want to clear `mutiSelectedObjects ` in the store to uncheck checkboxes after downlads

frontend/src/store/objectStore.ts
```diff
@@ -42,14 +43,14 @@ export const useObjectStore = defineStore('objectStore', () => {
     objectList.value = response.data;
   }
 
-  async function getObject(objectId: string) {
-    await objectService.getObject(objectId);
+  async function getObject(objectId: string, versionId?: string, download?: boolean) {
```
TimCsaky, line 46: not sure if you need the download parameter (?)

TimCsaky (approved):
I like the pattern!

frontend/tests/unit/views/HelloWorld.spec.ts
```diff
@@ -1,8 +1,9 @@
 import { mount } from '@vue/test-utils';
-import HelloWorld from '@/views/HelloWorld.vue';
+// import HelloWorld from '@/views/HelloWorld.vue';
 
+// Skip until a real view test can be added in here
 describe('HelloWorld.vue', () => {
-  it('renders', async () => {
+  it.skip('renders', async () => {
```
jujaga, line 6: At this point, I'd recommend just dropping this entire file.

jujaga (approved):
While I can't comment on the look and feel of what's been implemented, other than a few minor comments I think the structure seems reasonable.

frontend/tests/unit/views/HelloWorld.spec.ts
loneil: Will do

frontend/src/components/object/DeleteObjectButton.vue
```diff
@@ -0,0 +1,63 @@
...
+    required: true,
+  },
+  ids: {
+    type: Array<string>,
+    required: true,
+  },
+});
+
+// Deletion
+const displayNoFileDlg = ref(false);
```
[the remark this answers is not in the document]
loneil: 👍

frontend/src/components/object/DeleteObjectButton.vue
jujaga: We've got a whole host of soft-delete/hard-delete considerations we'll have to do at some point as well down the line. For now I think it's sufficient to just soft-delete and call it a day, as that's what's generally happening on S3Browser when you delete a file there anyways if I'm not mistaken.

frontend/src/components/object/DownloadObjectButton.vue
jujaga: It is true that this is an async function. However, the outcome of this function is fundamentally the invocation of a window.open call. As we don't really have any subsequent logic that depends on the resolution of window.open call, this looks sufficient.

frontend/src/store/objectStore.ts
jujaga: Yea I agree - with the function signature change on service layer, that download parameter doesn't need to exist anymore.

loneil: Indeed, only 1 get file mode at this point.

frontend/src/components/object/ObjectList.vue
loneil: fixed

frontend/src/components/object/DownloadObjectButton.vue
loneil: No it's just kicking them off and letting them download in their own time. I'll see about clearing those in my next PR when I do the delete too. That's a good point I think.

PR #12, opened 2022-12-20
Delete call

loneil:
# Description

Make api calls on delete button. At this point just loops as that's all COMS supports. Loop is done in store and the action takes an array. Doesn't do any DOM manipulation or window action so makes sense to me to have the loop here rather than in the component.

At this point delete doesn't look like it's doing anything. It does call the API with its functionality that is available at this point. so it's just writing a delete version and deleting the file in S3. Future changes to COMS and BCBox should filter the object list to not show deleted.

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/object/DeleteObjectButton.vue
+++ b/frontend/src/components/object/DeleteObjectButton.vue
@@ -8,6 +8,9 @@ import Button from 'primevue/button';
 import { useConfirm } from 'primevue/useconfirm';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 import Dialog from 'primevue/dialog';
+// State
+import { useObjectStore } from '@/store/objectStore';
+const objectStore = useObjectStore();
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/object/DeleteObjectButton.vue
```diff
@@ -25,16 +30,19 @@ const props = defineProps({
 
 // Deletion
 const displayNoFileDialog = ref(false);
-const confirmDelete = () => {
+const confirmDelete = async () => {
```
jujaga, line 33: Does this need to be declared as an async function? Looks like the only await in this scope is the await in the accept function, which is already in its own async closure.

loneil: Correct, not needed. Fixed

loneil:
@jujaga rebased and fixed comment

jujaga (approved):

kamorel (approved):

PR #13, opened 2022-12-22
Check user permissions before obtaining bucket list

kamorel:
# Description

Updates the bucket listing to include buckets where a user may not have explicit permissions at the bucket level, but has object permissions within it.

## Types of changes

## Checklist


frontend/src/store/bucketStore.ts
```diff
@@ -14,8 +21,19 @@ export const useBucketStore = defineStore('bucket', () => {
...
-      buckets.value = response.data;
+
+      const identityId = getKeycloak.value.tokenParsed?.idir_user_guid;
+      if (identityId) {
+        await userStore.searchUsers({ identityId: identityId });
+
+        if (userSearch.value.length) {
+          const userId = userSearch.value[0].userId;
+          const permResponse = (await bucketService.searchForPermissions({ userId, objectPerms: true })).data;
+          const uniqueIds = [...new Set(permResponse.map((x: any) => x.bucketId))];
```
jujaga, line 32: Can this be typecasted into x: Object or will it complain if we did that?

frontend/src/store/userStore.ts
```diff
@@ -1,38 +1,40 @@
...
   }
 
-  // Will 404. TODO DELETE
-  async function testBad() {
+  async function listIdps() {
     try {
       loading.value = true;
-      await userService.testBad();
+      const res = await userService.listIdps();
+      idps.value = res.data;
```
jujaga, line 29: Could oneline this into
```ts
idps.value = (await userService.listIdps()).data;
```

frontend/src/store/userStore.ts
```diff
@@ -1,38 +1,40 @@
...
     } catch (error) {
-      console.error(`bad error: ${error}`);
+      console.error(`listIdps error: ${error}`);
```
jujaga, line 31: We should look into some kind of frontend console logging library so that we aren't directly using console.* commands (this will be picked up by our linter soon)

frontend/src/store/bucketStore.ts
kamorel: It will complain that bucketId does not exist on type Object.

jujaga: Try embedding a direct in-line type then:
```ts
const uniqueIds = [...new Set(permResponse.map((x: { bucketId: string }) => x.bucketId))];
```

frontend/src/store/bucketStore.ts
```diff
@@ -14,8 +20,19 @@ export const useBucketStore = defineStore('bucket', () => {
   async function load() {
     try {
       loading.value = true;
-      const response = await bucketService.searchForBuckets();
-      buckets.value = response.data;
+
+      const identityId = getKeycloak.value.tokenParsed?.idir_user_guid;
+      if (identityId) {
+        await userStore.searchUsers({ identityId: identityId });
```
TimCsaky, line 26: i wonder if current user's coms `userId` should be saved to a property in the userStore or authStore to avoid having to do a coms user search search

frontend/src/store/userStore.ts
```diff
@@ -1,38 +1,40 @@
...
   const loading = ref(false);
 
-  async function listIdps() {
+  async function searchUsers(params: Object) {
     try {
       loading.value = true;
-      const res = await userService.listIdps();
-      idps.value = res.data;
+      const userResponse = await userService.searchForUsers(params);
+      userSearch.value = userResponse.data;
```
TimCsaky, line 15: guess this can be made more concise as with Jeremy's suggestion below

frontend/src/store/bucketStore.ts
jujaga: That should be the intent. On frontend bootup, assuming you're logged in, you should have in the user store at minimum your current identityId, but no knowledge of the COMS userId. If/when we do invoke the call to hydrate that piece of information, that should be written into the store, and actions that need that piece of information either pull that information from the store since its already there, or force it to do a user call to get it hydrated first. That way we can cache and reduce network calls.

frontend/src/store/bucketStore.ts
```diff
@@ -14,8 +20,19 @@ export const useBucketStore = defineStore('bucket', () => {
   async function load() {
     try {
       loading.value = true;
-      const response = await bucketService.searchForBuckets();
-      buckets.value = response.data;
+
+      const identityId = getKeycloak.value.tokenParsed?.idir_user_guid;
```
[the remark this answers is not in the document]
jujaga: This is going to have to turn into a utils level function - based on the incoming config change work in the other PR, this new utils function will have to iterate through all the defined properties to pull out the identityId from the tokenParsed object. This is fine for now for IDIR only situations, but this can't stay for long.

frontend/src/utils/constants.ts
```diff
@@ -3,6 +3,11 @@ export const KEYCLOAK = Object.freeze({
   REFRESH_TIME_MS: 10000,
 });
 
+export const IDENTITY_PROVIDERS = Object.freeze({
+  IDIR: 'idir',
+  BCEID_BASIC: 'bceid_basic',
```
TimCsaky, line 8: we'll need a third item:
`BCEID_BUSINESS: 'bceid_business'`
The field from the token we will use for identityId is also `bceid_user_guid`.

frontend/src/store/authStore.ts
```diff
@@ -16,10 +16,19 @@ export const useAuthStore = defineStore('auth', () => {
...
+
+  function getIdentityId() {
+    if (getIdentityProvider.value === IDENTITY_PROVIDERS.IDIR) {
+      return _keycloak.value.tokenParsed?.idir_user_guid;
+    }
+    // TODO: Confirm BCEID is correct
+    if (getIdentityProvider.value === IDENTITY_PROVIDERS.BCEID_BASIC) {
+      return _keycloak.value.tokenParsed?.bceid_user_guid;
+    }
+
```
TimCsaky, line 29: I beleieve we are allowing bceid business users to log in so we need another item:
You can add
```js
if (getIdentityProvider.value === IDENTITY_PROVIDERS.BCEID_BUSINESS) {
      return _keycloak.value.tokenParsed?.bceid_user_guid;
    }
```
The field from the token we will use for identityId is also `bceid_user_guid`.

TimCsaky (commented):
looks great. there's a minor addition to support bceid business users.. but it doesnt have to go in this pr

frontend/src/utils/constants.ts
TimCsaky: actually.. i wonder if there needs to be this 3rd user type in the /app/config/idplist-local.json file.
and I believe we were going to name that file idplist.json
and line 34 and 35 of this file 
should be changed in the ap.ts to :
```js
async function readIdpList(): Promise<string> {
  const idplistDefaultPath = './config/idplist.json';
  const idplistOverridePath = './config/idplist-local.json';
```

frontend/src/store/objectStore.ts
```diff
@@ -62,8 +62,10 @@ export const useObjectStore = defineStore('objectStore', () => {
         // TODO: Tags
 
         // Populate object
-        obj.metadata = metadataResponse.data[0];
-        obj.name = obj.metadata.metadata.find((x: any) => x.key === 'name')?.value;
+        if (metadataResponse.data.length && metadataResponse.data[0]) {
```
jujaga, line 65: Order of checking should be inverted. Checking for existence should always take precedence over checking an attribute of said object in order to best take advantage of short circuiting.
```js
if (metadataResponse.data[0] && metadataResponse.data.length) {
```

frontend/src/utils/constants.ts
```diff
@@ -3,6 +3,11 @@ export const KEYCLOAK = Object.freeze({
...
+});
+
```
jujaga, line 10: Let's revisit whether or not we really need to explicitly define this or not.

frontend/src/services/bucketService.ts
```diff
@@ -6,20 +6,39 @@ export default {
...
+   * @returns {Promise} An axios response
+   */
   searchForPermissions(params: Object) {
     return comsAxios().get(`${BUCKET_PERMISSION_PATH}`, { params });
   },
 
+  /**
+   * @function addPermissions
+   * Adds the given permissions to the bucket
+   * @param {Array<Object>} data Array containing permissions to add
```
jujaga, line 29: Missing bucketId @param definition

frontend/src/services/bucketService.ts
```diff
@@ -6,20 +6,39 @@ export default {
...
+   * @returns {Promise} An axios response
+   */
   addPermissions(bucketId: string, data: Array<Object>) {
     return comsAxios().put(`${BUCKET_PERMISSION_PATH}/${bucketId}`, data);
   },
 
+  /**
+ * @function deletePermission
+ * Deletes the given permission from the bucket
+ * @param {Object} params Object containing the permission to remove
```
jujaga, line 39: Missing bucketId @param definition

frontend/src/store/objectStore.ts
```diff
@@ -62,7 +62,7 @@ export const useObjectStore = defineStore('objectStore', () => {
         // TODO: Tags
 
         // Populate object
-        if (metadataResponse.data.length && metadataResponse.data[0]) {
+        if (metadataResponse.data[0] && metadataResponse.data.length) {
```
jujaga, line 65: Not completely sure, but might be able to get away with just one conditional check as the second conditional is redundant?
```ts
        if (metadataResponse.data[0]) {
```

frontend/src/utils/constants.ts
jujaga: Omitting busines bceid for now until we get more clarity on upstream IDP behavior.

app/config/idplist-local.json
```diff
@@ -10,5 +10,11 @@
     "idp": "bceid_basic",
     "identityKey": "bceid_user_guid",
     "searchable": false
+  },
+  {
+    "elevatedRights": false,
+    "idp": "bceid_business",
+    "identityKey": "bceid_business_guid",
```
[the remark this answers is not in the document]
jujaga: Omitting business bceid for now until we get more clarity on upstream IDP behavior.

jujaga (approved):

TimCsaky (approved):

PR #14, opened 2022-12-23
Backend Infrastructure cleanup

jujaga:
# Description

A bunch of eslint/testing fixes for backend. Make sure all your local files are in LF format, not CRLF format after this!

## Types of changes

## Checklist


frontend/package.json
```diff
@@ -43,7 +43,6 @@
     "eslint-plugin-prettier": "^4.2.1",
     "eslint-plugin-vue": "^9.3.0",
     "happy-dom": "^7.6.6",
-    "npm-run-all": "^4.1.5",
```
loneil, line 46: Looks like `build` uses **npm-run-all** (`run-p`) so removing this breaks the TS building part

kamorel (approved):

frontend/package.json
jujaga: Thanks for the catch. I'll take another look and see if we can rewrite the build or revert this change.

loneil (approved):

PR #15, opened 2022-12-23
Extending get/config endpoint to return idpList configuration

dc-bcgov:
# Description

Extending 'get/config' endpoint to return appropriate 'idpList' configuration, and append it to the 'frontend' config. Default will be used if override does not exist. File exist check is covered through import/catch, though if default is not defined error will be thrown.

## Checklist


app/app.ts
```diff
@@ -12,10 +12,12 @@ app.get('/', (req, res) => {
...
 // Frontend configuration endpoint
-app.get('/config', (_req, res, next) => {
+app.get('/config', async (_req, res, next) => {
   try {
-    const frontend = config.get('frontend');
-    res.status(200).json(frontend);
+    res.status(200).json({
+      ...config.get('frontend'),
+      idpList: await readIdpList()
+   });
```
jujaga, line 20: Fix indentation (missing a space).

jujaga (approved):

app/config/idplist-default.json
```diff
@@ -0,0 +1,14 @@
+[
+  {
+    "elevatedRights": true,
```
TimCsaky, line 3: wonder if `level: <int>` would work better (?)

app/app.ts
```diff
@@ -28,4 +30,12 @@ app.use('/favicon.ico', (_req, res) => {
 });
 app.use(staticFilesPath, express.static(path.join(__dirname, 'dist')));
 
+async function readIdpList(): Promise<string> {
+  const idplistDefaultPath = './config/idplist-default.json';
+  const idplistOverridePath = './config/idplist-override.json';
```
TimCsaky, line 35: would anyone prefer if overrides were in `idplist-local.json`

jujaga: Could change it to be named local instead - would match better with existing config library patterns. 👍

app/config/idplist-default.json
jujaga: We don't really have multi-tiered levels of access so I don't think we should go that route yet; a boolean suffices.

kamorel (approved):

PR #22, opened 2023-01-16
Feat/2970 bucket permission

dc-bcgov:
# Description

Created new 'BucketPermissionUser' component to support user search and adding bucket permissions. Added user will not be permanently saved in db unless one of the permission checkboxes is selected. Custom event 'cancel-bucket-permission-user' has been added to support parent-child communication i.e. to hide bucket permission user component.

## Checklist

### DIFF
--- a/frontend/src/components/bucket/BucketPermission.vue
+++ b/frontend/src/components/bucket/BucketPermission.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { onMounted, ref } from 'vue';
+import { onMounted, Ref, ref } from 'vue';
 import { storeToRefs } from 'pinia';
 import Button from 'primevue/button';
 import Checkbox from 'primevue/checkbox';
@@ -8,6 +8,7 @@ import DataTable from 'primevue/datatable';
 import { useBucketStore } from '@/store';
 import { Permissions } from '@/utils/constants';
 import BucketPermissionUser from './BucketPermissionUser.vue';
+import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 
 const props = defineProps<{
   bucketId: string;
@@ -32,7 +33,7 @@ const removeBucketUser = (userId: string) => {
   bucketStore.removeBucketUser(props.bucketId, userId);
 };
 
-const displayBucketPermissionUser : any = ref(false);
+const displayBucketPermissionUser : Ref<Boolean> = ref(false);
 
 const cancelBucketPermissionUser = () => {
   displayBucketPermissionUser.value = false;
--- a/frontend/src/components/bucket/BucketPermission.vue
+++ b/frontend/src/components/bucket/BucketPermission.vue
@@ -11,50 +11,71 @@ import { useBucketStore, useUserStore } from '@/store';
 import type { User } from '@/interfaces';
 import { Permissions } from '@/utils/constants';
 
-const showAddUser: any = ref(true);
-const addBucketPermissionUser = () => {
-  permissions.value.push(
-    {
-      userId: selectedUser.value.userId,
-      elevatedRights: true,
-      fullName: selectedUser.value.fullName,
-      create: false,
-      read: false,
-      update: false,
-      delete: false,
-      manage: false
-    }
-  );
-  resetSearchUsers();
-};
-
+// props
 const props = defineProps<{
   bucketId: string;
 }>();
 
+// stores
 const bucketStore = useBucketStore();
 const { loading, permissions } = storeToRefs(useBucketStore());
 
 const userStore = useUserStore();
 const { userSearch } = storeToRefs(useUserStore());
-const selectedUser = ref();
 
-const onInputSearchUsers = async (event: any) => {
-  if(event.target.value.length > 0) {
-    await userStore.searchUsers({ firstName: event.target.value });
-  } else if(event.target.value.length < 1) {
-    resetSearchUsers();
+// mounted
+onMounted(() => {
+  bucketStore.getBucketPermissions(props.bucketId);
+  resetBucketPermissionUsers();
+});
+
+//variables and methods
+const isShowAddUserBtn: any = ref(true);
+const isAddBPUBtnDisabled: any = ref(true);
+
+const selectedUser = ref();
+const onAddBucketPermissionUser = () => {
+  if(selectedUser.value !== null &&
+      !permissions.value.some(e => e.fullName === selectedUser.value.fullName)) {
+    permissions.value.push(
+      {
+        userId: selectedUser.value.userId,
+        elevatedRights: true,
+        fullName: selectedUser.value.fullName,
+        create: false,
+        read: false,
+        update: false,
+        delete: false,
+        manage: false
+      }
+    );
+    resetBucketPermissionUsers();
   }
 };
-const resetSearchUsers = () => {
+const onCancelBPU = () => {
+  isShowAddUserBtn.value = true;
+  isAddBPUBtnDisabled.value = true;
+  resetBucketPermissionUsers();
+};
+const onDropdownInput = async (event: any) => {
+  var inputValue = event.target.value;
+  if(inputValue.length > 0) {
+    await userStore.searchUsers({ search: event.target.value }); //TO DO: look at the bug with 'ma', 'sa'...
+  }
+  if(inputValue.length < 1) {
+    resetBucketPermissionUsers();
+  }
+  if (selectedUser.value !== null && selectedUser.value.fullName == undefined) {
+    isAddBPUBtnDisabled.value = true;
+  }
+};
+const onDropdownChange = () => {
+  isAddBPUBtnDisabled.value = selectedUser.value === null;
+};
+const resetBucketPermissionUsers = () => {
   userSearch.value = ([] as User[]);
   selectedUser.value = null;
 };
-
-onMounted(() => {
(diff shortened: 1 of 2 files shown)


frontend/src/components/bucket/BucketPermission.vue
```diff
@@ -30,17 +31,32 @@ const updateBucketPermission = (value: any, userId: string, permCode: string) =>
 const removeBucketUser = (userId: string) => {
   bucketStore.removeBucketUser(props.bucketId, userId);
 };
+
+const displayBucketPermissionUser : any = ref(false);
```
jujaga, line 35: Should this be of type boolean?

frontend/src/components/bucket/BucketPermissionUser.vue
```diff
@@ -0,0 +1,109 @@
...
+const showUserExistError: any = ref(false);
+const selectedUser = ref();
+
+const emit = defineEmits(['cancel-bucket-permission-user']);
+
+const cancelBucketPermissionUser = async () => {
+  isAddDisabled.value = true;
+  reset();
+  emit('cancel-bucket-permission-user');
+};
```
jujaga, line 28: There is no await in this callback function at all so this function declaration does not need to be async.

frontend/src/components/bucket/BucketPermissionUser.vue
```diff
@@ -0,0 +1,109 @@
...
+
+const onInput = async (event: any) => {
+  showUserExistError.value = ifUserExist();
+  var inputValue = event.target.value;
+  if(inputValue.length > 0) {
+    await userStore.searchUsers({ search: inputValue });
+  }
+  if(inputValue.length < 1) {
+    reset();
+  }
```
jujaga, line 42: This conditional can be condensed to take advantage of general JS truthiness behavior. For example, the following ternary achieves the same effect:
```ts
inputValue.length ? await userStore.searchUsers({ search: inputValue }) : reset();
```

frontend/src/components/bucket/BucketPermissionUser.vue
```diff
@@ -0,0 +1,109 @@
...
+  emit('cancel-bucket-permission-user');
+};
+
+const ifUserExist = () => {
+  return permissions.value.some(e => e.fullName === selectedUser.value.fullName);
+};
+
+const onInput = async (event: any) => {
+  showUserExistError.value = ifUserExist();
+  var inputValue = event.target.value;
```
jujaga, line 36: Never use var. Variables must be declared as either a const, or a let if there is intentional reassignment going on.

frontend/src/components/bucket/BucketPermissionUser.vue
```diff
@@ -0,0 +1,109 @@
...
+        userId: selectedUser.value.userId,
+        elevatedRights: true,
+        fullName: selectedUser.value.fullName,
+        create: false,
+        read: false,
+        update: false,
+        delete: false,
+        manage: false
+      }
+    );
```
jujaga, line 66: Style: Not too important, but in situations where there's just an object declaration for the body, we can remove unnecessary indentation by having the curly braces trail the parentheses instead:
```ts
    permissions.value.push({
      userId: selectedUser.value.userId,
      elevatedRights: true,
      fullName: selectedUser.value.fullName,
      create: false,
      read: false,
      update: false,
      delete: false,
      manage: false
    });
```
Also reduces the amount of lines used in general.

frontend/src/components/bucket/BucketPermission.vue
```diff
@@ -30,17 +31,32 @@ const updateBucketPermission = (value: any, userId: string, permCode: string) =>
...
+
+const cancelBucketPermissionUser = () => {
+  displayBucketPermissionUser.value = false;
+};
 </script>
 
 <template>
   <div>
-    <Button class="mt-1 mb-4">
+    <Button
```
TimCsaky, line 44: i wonder if you want this button to be in the `BucketPermissionUser` component template, and have the show/hide logic in that component to keep it completely contained and might avoid some of the emitting. That might not be possible but maybe something to consider.

frontend/src/components/bucket/BucketPermissionUser.vue
```diff
@@ -0,0 +1,109 @@
...
+
+const ifUserExist = () => {
+  return permissions.value.some(e => e.fullName === selectedUser.value.fullName);
+};
+
+const onInput = async (event: any) => {
+  showUserExistError.value = ifUserExist();
+  var inputValue = event.target.value;
+  if(inputValue.length > 0) {
+    await userStore.searchUsers({ search: inputValue });
```
TimCsaky: i cant remember if we are searching by just email address (I think so) in which case maybe this could be something like `userStore.searchUsers({ email: inputValue })` (?) i guess i depends on the searchUser() function.

frontend/src/components/bucket/BucketPermissionUser.vue
```diff
@@ -0,0 +1,109 @@
...
+};
+
+const ifUserExist = () => {
+  return permissions.value.some(e => e.fullName === selectedUser.value.fullName);
+};
+
+const onInput = async (event: any) => {
+  showUserExistError.value = ifUserExist();
+  var inputValue = event.target.value;
+  if(inputValue.length > 0) {
```
TimCsaky: maybe this where we could only call searchUsers if character length is > 3

frontend/src/components/bucket/BucketPermission.vue
dc-bcgov: This is the old code. Should I change it?

frontend/src/components/bucket/BucketPermissionUser.vue
TimCsaky: it might be something to tackle in another PR but i think we talked about filtering searchUsers to `&idp=idir` unless inputValue is a full email address

frontend/src/components/bucket/BucketPermissionUser.vue
```diff
@@ -0,0 +1,109 @@
...
+const emit = defineEmits(['cancel-bucket-permission-user']);
+
+const cancelBucketPermissionUser = async () => {
+  isAddDisabled.value = true;
+  reset();
+  emit('cancel-bucket-permission-user');
+};
+
+const ifUserExist = () => {
+  return permissions.value.some(e => e.fullName === selectedUser.value.fullName);
```
TimCsaky, line 31: i think we need to reference users by either userId or email address to avoid finding duplicates

frontend/src/components/bucket/BucketPermissionUser.vue
```diff
@@ -0,0 +1,109 @@
...
+  showUserExistError.value = ifUserExist();
+  isAddDisabled.value = (selectedUser.value === null || ifUserExist());
+};
+
+const onAdd = () => {
+  if(selectedUser.value !== null && !ifUserExist()) {
+    permissions.value.push(
+      {
+        userId: selectedUser.value.userId,
+        elevatedRights: true,
```
TimCsaky: maybe somethingf ro a later PR (?) i think we need to get elevatedRights value by checking the config for their idp.
something like `config.value.idpList.find((idp: any) => idp.idp === userSearch.value[0].idp)?.elevatedRights ?? false;`

TimCsaky (commented):
looks good. I guess you figured out a lot of Vue stuff already which is great.
to be honest I havent run this locally to test.
My comments mostly relate to if/how we search for users.
maybe something to discuss in a future PR

frontend/src/components/bucket/BucketPermission.vue
dc-bcgov: Yeah, that would be better (pls correct me if wrong) but it seems like emit is the new way of two-way binding, props is parent-child only.

jujaga: It's worth considering if this kind of behavior appears in multiple places.

frontend/src/components/bucket/BucketPermission.vue
```diff
@@ -1,12 +1,14 @@
...
-import { onMounted } from 'vue';
+import { onMounted, Ref, ref } from 'vue';
 import { storeToRefs } from 'pinia';
 import Button from 'primevue/button';
 import Checkbox from 'primevue/checkbox';
 import Column from 'primevue/column';
 import DataTable from 'primevue/datatable';
 import { useBucketStore } from '@/store';
 import { Permissions } from '@/utils/constants';
+import BucketPermissionUser from './BucketPermissionUser.vue';
```
kamorel, line 10: Use @ imports where possible, and alphabetize import order. '@/components/bucket/BucketPermissionUser.vue' So this would go above @/store

frontend/src/components/bucket/BucketPermission.vue
```diff
@@ -30,17 +32,32 @@ const updateBucketPermission = (value: any, userId: string, permCode: string) =>
 const removeBucketUser = (userId: string) => {
   bucketStore.removeBucketUser(props.bucketId, userId);
 };
+
+const displayBucketPermissionUser : Ref<Boolean> = ref(false);
```
kamorel, line 36: Style: Extra space after displayBucketPermissionUser

kamorel (changes requested):

kamorel (approved):

jujaga (approved):

PR #23, opened 2023-01-16
Object Permissions

loneil:
# Description
Allow users to set object permissions on files they have manage permissions for. The "Add User" part is on an upcoming ticket.

image

Don't show the actions on a file that you don't have the corresponding permission for.

Additionally a fix for
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3012

## Types of changes

## Checklist

## Further comments

There's probably some efficiencies that can be done here. This is mostly the same code as the bucket permissions screen and table, but with enough changes I couldn't quickly figure out how to generalize it. There's some differences between the 2 things. 
Possibly could be refactored to a common component and/or composable, but need to be sure the behavior would stay the same.

At the very least the component SCSS should be extracted out so that it's not duplicated. Would need to decide if above path would be looked at, if not, move to global scss for this type of checkbox table.

### DIFF
--- a/frontend/src/components/object/ObjectList.vue
+++ b/frontend/src/components/object/ObjectList.vue
@@ -49,7 +49,6 @@ const listObjects = async () => {
       bucketStore.getBucketPermissionsForUser(route.query.bucketId?.toString() || ''),
       objectStore.listObjects({ bucketId: route.query.bucketId })
     ]);
-    await objectStore.listObjects({ bucketId: route.query.bucketId });
   } catch (error: any) {
     toast.add({ severity: 'error', summary: 'Unable to load Objects.', detail: error, life: 5000 });
   }
--- a/frontend/src/components/object/ObjectList.vue
+++ b/frontend/src/components/object/ObjectList.vue
@@ -9,6 +9,7 @@ import Button from 'primevue/button';
 import { useToast } from 'primevue/usetoast';
 // State
 import { storeToRefs } from 'pinia';
+import { useBucketStore } from '@/store';
 import { useObjectStore } from '@/store';
 // Components
 import DeleteObjectButton from './DeleteObjectButton.vue';
@@ -17,6 +18,7 @@ import ObjectSidebar from './ObjectSidebar.vue';
 import ObjectTable from './ObjectTable.vue';
 import ObjectUpload from './ObjectUpload.vue';
 
+const bucketStore = useBucketStore();
 const objectStore = useObjectStore();
 const { multiSelectedObjects } = storeToRefs(objectStore);
 const route = useRoute();
@@ -43,6 +45,10 @@ const closeUpload = () => {
 
 const listObjects = async () => {
   try {
+    await Promise.all([
+      bucketStore.getBucketPermissionsForUser(route.query.bucketId?.toString() || ''),
+      objectStore.listObjects({ bucketId: route.query.bucketId })
+    ]);
     await objectStore.listObjects({ bucketId: route.query.bucketId });
   } catch (error: any) {
     toast.add({ severity: 'error', summary: 'Unable to load Objects.', detail: error, life: 5000 });
@@ -108,6 +114,4 @@ onMounted(() => {
   </div>
 </template>
 
-<style scoped>
-
-</style>
+<style scoped></style>
--- a/frontend/src/components/object/ObjectPermission.vue
+++ b/frontend/src/components/object/ObjectPermission.vue
@@ -22,8 +22,12 @@ const objectStore = useObjectStore();
 const { loading, selectedObjectPermissions } = storeToRefs(useObjectStore());
 
 onMounted(() => {
-  useToaster(() => objectStore.getObjectPermissions(props.objectId), { summary: 'Unable to load permissions.' });
+  fetchPermissions();
 });
+
+const fetchPermissions = () => {
+  useToaster(() => objectStore.getObjectPermissions(props.objectId), { summary: 'Unable to load permissions.' });
+};
 
 const updateObjectPermission = (value: any, userId: string, permCode: string) => {
   if (value) {
@@ -33,8 +37,9 @@ const updateObjectPermission = (value: any, userId: string, permCode: string) =>
   }
 };
 
-const removeBucketUser = (userId: string) => {
-  objectStore.removeObjectUser(props.objectId, userId);
+const removeObjectUser = async (userId: string) => {
+  await objectStore.removeObjectUser(props.objectId, userId);
+  fetchPermissions();
 };
 </script>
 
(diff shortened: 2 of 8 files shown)


frontend/src/components/object/ObjectPermission.vue
```diff
@@ -0,0 +1,172 @@
...
+:deep(.truncate) {
+  max-width: 1px;
+  white-space: nowrap;
+
+  > div {
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+}
+</style>
```
loneil, line 172: This CSS is the same as the bucket one. Possibly this whole table component could be generalized...
At the least extract the CSS at some point

frontend/src/components/object/ObjectList.vue
```diff
@@ -43,7 +45,10 @@ const closeUpload = () => {
 
 const listObjects = async () => {
   try {
-    await objectStore.listObjects({ bucketId: route.query.bucketId });
+    await Promise.all([
+      bucketStore.getBucketPermissionsForUser(route.query.bucketId?.toString() || ''),
+      objectStore.listObjects({ bucketId: route.query.bucketId })
```
loneil, line 50: So if you directly navigate to a object list landing page, you will have your perms for that bucket populated.

frontend/src/services/objectService.ts
```diff
@@ -108,5 +109,38 @@ export default {
...
+
+  /**
+   * @function searchForPermissions
+   * Returns a list of permissions for the object
+   * @param {Object} params Optional object containing the data to filter against
+   * @returns {Promise} An axios response
+   */
+  searchForPermissions(objectId: string, params?: Object) {
+    return comsAxios().get(`${OBJECT_PERMISSION_PATH}/${objectId}`, { params: params });
+  },
```
jujaga, line 144: We have a permissionService.ts file - this logic should reside there instead of this file.

frontend/src/services/permissionService.ts
```diff
@@ -46,4 +46,15 @@ export default {
...
     return comsAxios().get(`${PATH}/${OBJECT}`, { params: params });
   },
+
+  /**
+     * @function objectGetPermissions
+     * Returns a list of permissions for the object
+     * @param {string} objectId ID of the object
+     * @param {Object} params Optional object containing the data to filter against
+     * @returns {Promise} An axios response
+     */
```
jujaga, line 56: JSDocs are overindented by 2 spaces

frontend/src/store/objectStore.ts
```diff
@@ -103,15 +111,104 @@ export const useObjectStore = defineStore('objectStore', () => {
...
+  async function removeObjectUser(bucketId: string, userId: string) {
+    try {
+      loading.value = true;
+
+      for (const [, value] of Object.entries(Permissions)) {
+        await objectService.deletePermission(bucketId, {
+          userId,
+          permCode: value,
+        });
+      }
```
jujaga, line 190: Two thoughts:
1) If we're only using the values and don't need the key/value pairing, we should probably be using `Object.values()` instead.
```ts
for (const value of Object.values(Permissions)) {
```
2) Since this looks like a general async dispatcher pattern, it may be better to wrap this into a `Promise.all()` for synchronization purposes:
```ts
await Promise.all(
  Object.values(Permissions)
    .map(perm => objectService.deletePermission(bucketId, {
      userId,
      permCode: value,
    }));
);
```

jujaga (commented):
Semantically looks good. Just a few comments for consideration.

frontend/src/components/object/ObjectTable.vue
```diff
@@ -1,132 +1,196 @@
...
-      <template #body="{ data }">
-        <div
-          v-if="data.id?.length > 15"
-          v-tooltip.bottom="{ value: data.id }"
-        >
-          {{ data.id }}
+      <!-- eslint-enable vue/no-v-model-argument -->
+      <template #header>
+        <div class="flex">
+          <font-awesome-icon
```
TimCsaky, line 181: possibly the icon change here (?)

frontend/src/store/objectStore.ts
```diff
@@ -82,8 +86,12 @@ export const useObjectStore = defineStore('objectStore', () => {
             obj.metadata = metadata;
             obj.name = metadata.metadata.find((x: { key: string }) => x.key === 'name')?.value;
           }
-        });
 
+          // Add the permissions to each object list item
```
TimCsaky, line 90: i think in line 77 we are removing the perms and then adding them again here in line 91. not sure if that could be optimized.

TimCsaky (commented):
a lot of good stuff in here!

dc-bcgov (approved):

TimCsaky:
just ran this in the browser and it's looking good.
couple more minor things we will probably have to look at later:
- i was able to remove myself (remove user) from the list of users with permissions. But then i still see the file in objects table. i think i still had the read permission before i clicked the remove button.
- there's some kind of flickering when the objects table is refreshed, i wonder if there's a black or missing css background somewhere.

frontend/src/store/objectStore.ts
loneil: I think the line 77 being referred to is just stashing the specific IDs from the permissions to fetch the objects? So not removing them?
(file has changed since so line 77 might be something else from original comment)

frontend/src/components/object/ObjectTable.vue
loneil: Fixed

frontend/src/components/object/ObjectTable.vue
```diff
@@ -1,132 +1,196 @@
...
+            v-if="isActionAllowed(data.permissions, Permissions.READ)"
+            :mode="ButtonMode.ICON"
+            :ids="[data.id]"
+          />
+          <Button
+            v-if="isActionAllowed(data.permissions, Permissions.MANAGE)"
+            class="p-button-lg p-button-rounded p-button-text"
+            @click="showPermissions(data.id, data.name)"
+          >
+            <font-awesome-icon icon="fa-solid fa-gear" />
```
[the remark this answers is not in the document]
loneil: Fixed

frontend/src/services/objectService.ts
loneil: This one actually unused, removing

frontend/src/store/objectStore.ts
loneil: Used the same logic as in existing bucketStore removeBucketUser.
Think there's possibly a way to take in the service method and have this permission logic extracted to a common util function (for buckets and objects). So someone could take a look at that at some point and implement your logic above for both (ideally in one palce)

jujaga (approved):

frontend/src/store/objectStore.ts
TimCsaky: my point was: we are looping through permResponse in line 77, the perms can be added there instead of doing another loop. only mention this because for a bucket full of objects this could impact the speed

PR #32, opened 2023-02-01
Add user search to object permissions modal

kamorel:
# Description

1. Adds user search to object permissions modal.
2. SearchUsers now accepts any UserPermissions array
3. Realigns table headers in bucket/object perm tables.
4. Removes scoped styling and relies on main.css file.
5. Updates bucket permissions function to get unique users list in a single API call

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/assets/main.scss
+++ b/frontend/src/assets/main.scss
@@ -16,6 +16,14 @@
   }
 }
 
+.content-center {
+  text-align: center !important;
+}
+
 .content-right {
   text-align: right !important;
+}
+
+.permissions-modal {
+  width: 800px;
 }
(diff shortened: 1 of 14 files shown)


jujaga (approved):

frontend/src/assets/main.scss
```diff
@@ -16,6 +16,14 @@
   }
 }
 
+.content-center {
```
TimCsaky, line 19: you probably knew but primeflex (installed) has some built in css utility classes like `text-align` and `text-justify`. 
but i assume here we want to override into child components using !important.

frontend/src/assets/main.scss
```diff
@@ -16,6 +16,14 @@
...
+  text-align: center !important;
+}
+
 .content-right {
   text-align: right !important;
 }
+
+.permissions-modal {
+  width: 800px;
```
TimCsaky, line 28: this seems risky. might be better to do something like `max-width: 800px; width: 80%`

TimCsaky (approved):

PR #35, opened 2023-02-10
Add missing information to object details screens

kamorel:
# Description

- Adds bucket information to object properties screen
- Adds action buttons to object properties screen
- Sidebar object details no longer displays all the information
- Adds back the 'Manage by' information

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/form/GridRow.vue
+++ b/frontend/src/components/form/GridRow.vue
@@ -1,0 +1,42 @@
+<script setup lang="ts">
+defineProps({
+  label: {
+    type: String,
+    default: undefined
+  },
+  value: {
+    type: String,
+    default: undefined
+  },
+  link: {
+    type: Object,
+    default: undefined
+  }
+});
+
+</script>
+
+<template>
+  <div
+    v-if="value && label"
+    class="col-3"
+  >
+    {{ label }}:
+  </div>
+  <div
+    v-if="value && link"
+    class="col-9"
+  >
+    <router-link
+      :to="link"
+    >
+      {{ value }}
+    </router-link>
+  </div>
+  <div
+    v-if="value && !link"
+    class="col-9"
+  >
+    {{ value }}
+  </div>
+</template>
--- a/frontend/src/components/form/GridRow.vue
+++ b/frontend/src/components/form/GridRow.vue
@@ -13,7 +13,6 @@ defineProps({
     default: undefined
   }
 });
-
 </script>
 
 <template>
--- a/frontend/src/components/object/ObjectAccess.vue
+++ b/frontend/src/components/object/ObjectAccess.vue
@@ -4,6 +4,7 @@ import { storeToRefs } from 'pinia';
 import { useUserStore } from '@/store';
 import { Permissions } from '@/utils/constants';
 
+import type { Ref } from 'vue';
 import type { Permission } from '@/interfaces';
 
 const userStore = useUserStore();
@@ -16,7 +17,7 @@ const props = defineProps({
   }
 });
 
-const managedBy = ref();
+const managedBy: Ref<string | undefined> = ref();
 
 async function getUserData() {
   const uniqueIds = [...new Set(props.objectAccess?.permissions
--- a/frontend/src/components/object/ObjectAccess.vue
+++ b/frontend/src/components/object/ObjectAccess.vue
@@ -1,12 +1,40 @@
-<!-- <script setup lang="ts">
+<script setup lang="ts">
+import { onMounted, ref, watch} from 'vue';
+import { storeToRefs } from 'pinia';
+import { useUserStore } from '@/store';
+import { Permissions } from '@/utils/constants';
 
-defineProps({
+import type { Permission } from '@/interfaces';
+
+const userStore = useUserStore();
+const { userSearch } = storeToRefs(userStore);
+
+const props = defineProps({
   objectAccess: {
     type: Object,
     default: undefined
   }
 });
 
+const managedBy = ref();
+
+async function getUserData() {
+  const uniqueIds = [...new Set(props.objectAccess?.permissions
+    .filter( (x: Permission) => x.permCode === Permissions.MANAGE )
+    .map((x: Permission) => x.userId))];
+
+  await userStore.searchUsers({userId: uniqueIds} );
+
+  managedBy.value = userSearch.value.map( x => x.fullName ).join( ', ');
+}
+
+onMounted(() => {
+  getUserData();
+});
+
+watch( props, async () => {
+  await getUserData();
+});
 </script>
 
 <template>
@@ -20,7 +48,7 @@ defineProps({
       Managed by:
     </div>
     <div class="col-9">
-      {{ objectAccess?.managedBy }}
+      {{ managedBy }}
     </div>
   </div>
-</template> -->
+</template>
--- a/frontend/src/components/object/ObjectAccess.vue
+++ b/frontend/src/components/object/ObjectAccess.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { onMounted, ref, watch} from 'vue';
+import { onMounted, ref, watch } from 'vue';
 import { storeToRefs } from 'pinia';
 import { useUserStore } from '@/store';
 import { Permissions } from '@/utils/constants';
@@ -11,7 +11,7 @@ const userStore = useUserStore();
 const { userSearch } = storeToRefs(userStore);
 
 const props = defineProps({
-  objectAccess: {
+  objectInfo: {
     type: Object,
     default: undefined
   }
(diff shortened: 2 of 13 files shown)


frontend/src/components/object/ObjectAccess.vue
```diff
@@ -1,12 +1,41 @@
-<!-- <script setup lang="ts">
+<script setup lang="ts">
+import { onMounted, ref, watch} from 'vue';
```
jujaga, line 2: Style: Missing space between watch and }

frontend/src/components/object/ObjectAccess.vue
```diff
@@ -1,12 +1,41 @@
...
     type: Object,
     default: undefined
   }
 });
 
+const managedBy: Ref<string | undefined> = ref();
+
+async function getUserData() {
+  const uniqueIds = [...new Set(props.objectAccess?.permissions
+    .filter( (x: Permission) => x.permCode === Permissions.MANAGE )
```
jujaga, line 24: Style: Inconsistent spacing between parentheses. Do a VSCode format right click format here

frontend/src/components/object/ObjectAccess.vue
```diff
@@ -1,12 +1,41 @@
...
+  managedBy.value = userSearch.value.map( x => x.fullName ).join( ', ');
+}
+
+onMounted(() => {
+  getUserData();
+});
+
+watch( props, async () => {
+  await getUserData();
+});
```
jujaga, line 38: If I understand this correctly, this registers an event hook handler when props changes (assuming it detects deep object reassignments). While the async and await thing does make sense, in this case, it may be better to omit that as events are already asynchronously handled by nature of being triggered by an event. There's also nothing in the code that needs to explicitly block on the completion of getUserData, so the await (and async annotation on the callback) isn't actually needed here.

frontend/src/components/object/ObjectFileDetails.vue
```diff
@@ -2,25 +2,42 @@
 import { onMounted, ref } from 'vue';
 import { useRoute } from 'vue-router';
 import { storeToRefs } from 'pinia';
+import Button from 'primevue/button';
```
jujaga, line 5: Style: We'll want to reorder the imports so that the pinia one comes before the vue one.

frontend/src/components/object/ObjectFileDetails.vue
```diff
@@ -2,25 +2,42 @@
...
+  ObjectProperties,
+  ObjectTag
+} from '@/components/object';
+
+import { useObjectStore, useUserStore } from '@/store';
+import { Permissions } from '@/utils/constants';
 
 import type { Ref } from 'vue';
 import type { COMSObject } from '@/interfaces';
+import { ButtonMode } from '@/interfaces/common/enums';
```
jujaga, line 24: Style: Worthwhile changing this to import type to preserve consistency assuming the compiler doesn't complain.

frontend/src/components/object/ObjectFileDetails.vue
```diff
@@ -2,25 +2,42 @@
...
+const { currentUser } = storeToRefs(useUserStore());
+
 const objectInfo: Ref<COMSObject> = ref({} as COMSObject);
 const isInfoLoaded: Ref<Boolean> = ref(false);
-const { objectList } = storeToRefs(useObjectStore());
 
-const getObjectInfo = async (objId: string) => {
+const permissionsVisible = ref(false);
+const permissionsObjectId = ref('');
+const permissionsObjectName = ref('');
```
jujaga, line 38: TS Style: We'll likely want to be explicit about the Ref template typings here (Ref<boolean>, Ref<string>, etc)

frontend/src/components/object/ObjectList.vue
```diff
@@ -6,11 +6,12 @@ import { storeToRefs } from 'pinia';
...
-import DownloadObjectButton from '@/components/object/DownloadObjectButton.vue';
-import ObjectSidebar from '@/components/object/ObjectSidebar.vue';
-import ObjectTable from '@/components/object/ObjectTable.vue';
-import ObjectUpload from '@/components/object/ObjectUpload.vue';
+import {
+  DeleteObjectButton,
+  DownloadObjectButton,
+  ObjectSidebar,
+  ObjectTable,
+  ObjectUpload} from '@/components/object';
```
jujaga, line 14: Style: There should be a newline between ObjectUpload and }

frontend/src/components/object/ObjectProperties.vue
```diff
@@ -1,13 +1,55 @@
 <script setup lang="ts">
+import { onMounted, ref, watch } from 'vue';
+import { storeToRefs } from 'pinia';
+import { useBucketStore, useUserStore } from '@/store';
+import GridRow from '@/components/form/GridRow.vue';
+import { RouteNames } from '@/utils/constants';
```
jujaga, line 6: Style: Import ordering should be done based on library name. Newline between external and internal deps.
```ts
import { onMounted, ref, watch } from 'vue';

import GridRow from '@/components/form/GridRow.vue';
import { useBucketStore, useUserStore } from '@/store';
import { RouteNames } from '@/utils/constants';
```

frontend/src/components/object/ObjectProperties.vue
```diff
@@ -1,13 +1,55 @@
...
+
+onMounted(() => {
+  getBucketData();
+  getUserData();
+});
+
+watch( props, async () => {
+  await getBucketData();
+  await getUserData();
+});
```
jujaga, line 51: Async/await: Is there an actual linear dependency where getUserData needs to wait on getBucketData to complete first? If not, we should let these both fire asynchronously. That means the following is sufficient:
```ts
watch(props, () => {
  getBucketData();
  getUserData();
});
```

frontend/src/components/object/ObjectUpload.vue
```diff
@@ -1,15 +1,12 @@
...
 import { ref } from 'vue';
 import { useRoute } from 'vue-router';
-// PrimeVue
 import { useToast } from 'primevue/usetoast';
 import Button from 'primevue/button';
 import FileUpload from 'primevue/fileupload';
-// State
-import { useObjectStore } from '@/store';
-// Component
+
```
jujaga, line 7: Style: We should try to reorder the primevue/* imports to come before the vue ones now that we're removing the comment groupings on them unless that causes a compilation behavior due to import ordering.

jujaga (commented):
Looking pretty good from what I can gather; mostly style fixes and a few minor tweaks needed. 👍

frontend/src/components/object/ObjectFileDetails.vue
```diff
@@ -29,32 +46,88 @@ const getObjectInfo = async (objId: string) => {
...
+          :ids="[objectInfo.id]"
+        />
+      </div>
+    </div>
+
+    <div
+      v-if="isInfoLoaded"
+      class="pl-2"
+    >
+      <ObjectProperties :object-properties="objectInfo" />
```
TimCsaky, line 100: you could rename these props to `object-info="objectInfo"` so that they are received in the child component as 
`objectInfo`

frontend/src/components/object/ObjectFileDetails.vue
```diff
@@ -2,25 +2,42 @@
...
 const objectInfo: Ref<COMSObject> = ref({} as COMSObject);
 const isInfoLoaded: Ref<Boolean> = ref(false);
-const { objectList } = storeToRefs(useObjectStore());
 
-const getObjectInfo = async (objId: string) => {
+const permissionsVisible = ref(false);
+const permissionsObjectId = ref('');
+const permissionsObjectName = ref('');
+
+const getObjectInfo = (objId: string) => {
```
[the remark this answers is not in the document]
TimCsaky: separate note. Does this component call the objectStore.listObjects(objId) method.. in case someone navigates straight to this page. Rather than just checking the existing objectList.value.
maybe Luke was going to take a look at this.. or it;s for another ticket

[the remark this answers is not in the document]
loneil: I have a fix for that almost to go, but this ticket will merge conflict the heck out of that
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3049

frontend/src/components/object/ObjectProperties.vue
loneil: What's the need for the watches in these components. It's a load->loading->display pattern right? Or are there refreshes somewhere or subscriptions that would update that data?

frontend/src/components/object/ObjectFileDetails.vue
kamorel: As discussed - enums can't be imported as a type.

frontend/src/components/object/ObjectProperties.vue
kamorel: For the sidebar in the object list, the first time it is opened it gets mounted but if you click the info icon again on a different object it doesn't remount, it just receives new props.

frontend/src/components/form/GridRow.vue
```diff
@@ -0,0 +1,41 @@
...
+    v-if="value && label"
+    class="col-3"
+  >
+    {{ label }}:
+  </div>
+  <div
+    v-if="value && link"
+    class="col-9"
+  >
+    <router-link
```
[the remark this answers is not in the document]
kamorel: I agree with you here, but unfortunately nested div's break the formatting for some reason I wasn't able to figure out.

jujaga (approved):

PR #36, opened 2023-02-14
Implementation of 2966 bucket configure

dc-bcgov:
This pr contains implementation of Bucket Configure feature. It includes 'Configure New Bucket' dialog and form as a separate components. Fields on form are generated by common TextInput component. TextInput supports field validation (required at the moment). The submit of the form is not allowed unless all the fields are populated. Update Bucket Configure is happening on wheel icon click and brings up the same dialog, but it loads the updating bucket into the fields instead of being empty.

## Checklist


loneil:
Discussed REDACTED field password write issue in Discord.

loneil:
Could also consider validating bucket url to ensure it's URL
https://vee-validate.logaretm.com/v4/guide/global-validators#url

image

frontend/src/components/bucket/BucketList.vue
```diff
@@ -1,14 +1,28 @@
...
+const displayBucketConfig: Ref<boolean> = ref(false);
+
+const bucketConfigHeader: Ref<string> = ref('');
+
+const bucketConfigTitle: Ref<string> = ref('');
+
+const isBucketConfigUpdate: Ref<boolean> = ref(false);
+
+const bucketToUpdate: Ref<Bucket> = ref({} as Bucket);
+
```
kamorel, line 25: Remove newlines between declarations.

frontend/src/components/bucket/BucketConfigDialog.vue
```diff
@@ -0,0 +1,50 @@
...
+  display: boolean;
+  header: string;
+  title: string;
+  isUpdate: boolean;
+  bucket?: Bucket;
+}>();
+
+const emit = defineEmits(['close-bucket-config']);
+
+const closeDialog = async () => {
```
kamorel, line 16: I don't believe this async is necessary as we are not awaiting anything here.

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -0,0 +1,139 @@
...
+        severity: 'error',
+        summary: 'Bucket configuration was not successful',
+        detail: error,
+        life: 5000
+      }
+    );
+  }
+};
+
+const onCancel = async () => {
```
kamorel, line 78: Same as above. Not awaiting anything so probably dont need this async.

kamorel:
> Could also consider validating bucket url to ensure it's URL https://vee-validate.logaretm.com/v4/guide/global-validators#url
Valid, but i'd prefer to have the FE validation schema match that of COMS which it is currently doing.

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -0,0 +1,139 @@
...
+  try {
+    const formBucket = {
+      bucketName: values.bucketName,
+      bucket: values.bucket,
+      endpoint: values.endpoint,
+      accessKeyId: values.accessKeyId,
+      secretAccessKey: values.secretAccessKey,
+      key: values.key
+    } as Bucket;
+
```
kamorel, line 49: Use object spread syntax.
const formBucket = { ...values } as Bucket;

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -0,0 +1,139 @@
...
+    props.isUpdate ?
+      await bucketStore.updateBucket(props.bucket!.bucketId, formBucket) :
```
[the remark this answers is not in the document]
TimCsaky: Maybe when updating the config you can do:
- use the password field type (without the show/hide eye icon)
- You could have the field value as 'REDACTED' but it will show with dots
- if a user clicks on field, the input is cleared (@focus event)
- if user wants to update it they can type in the new value but it will show as dots
- on submit, include field value to formBucket object only if it is not empty and not 'REDACTED'

frontend/src/components/bucket/BucketList.vue
```diff
@@ -54,4 +107,10 @@ h1 {
 h3 {
   font-weight: bold;
 }
+
+button {
```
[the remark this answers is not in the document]
dc-bcgov: Found for margin only.

frontend/src/components/bucket/BucketConfigForm.vue
dc-bcgov: Kyle, here lint forces me to define the rest of the object. For example: region cannot be empty (getting error from the back-end) and don't have this info initialy.

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -0,0 +1,154 @@
...
+});
+
+const onSubmit = async (values: any) => {
+
+  try {
+    const formBucket = {
+      bucketName: values.bucketName,
+      bucket: values.bucket,
+      endpoint: values.endpoint,
+      accessKeyId: values.accessKeyId !== 'REDACTED' && values.accessKeyId !== '' ?
```
TimCsaky, line 46: having noticed that these two fields can't be empty (due to validation rules).. i guess you can just do:
`accessKeyId: values.accessKeyId !== 'REDACTED' ? values.accessKeyId : undefined`

frontend/src/components/bucket/BucketConfigDialog.vue
```diff
@@ -0,0 +1,50 @@
...
+        </div>
+      </template>
+      <BucketConfigForm
+        :is-update="props.isUpdate"
+        :bucket="props.bucket"
+        @submit-bucket-config="closeDialog"
+        @cancel-bucket-config="closeDialog"
+      />
+    </Dialog>
+  </div>
```
jujaga, line 49: Unnecessary div wrapper - as it only wraps one DOM element, we can remove it and instead have the Dialog element be the top level template element instead.
Also generally wondering whether we need to be breaking down this dialog into its own component or not as this isn't a very reusable component and very specific to the bucket context.

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -0,0 +1,154 @@
...
+
+const props = defineProps<{
+  isUpdate: boolean;
+  bucket?: Bucket;
+}>();
+
+const bucketStore = useBucketStore();
+const toast = useToast();
+
+const emit = defineEmits(['submit-bucket-config', 'cancel-bucket-config']);
```
jujaga, line 19: Alphabetically order emit attribute strings.

frontend/src/components/bucket/BucketList.vue
```diff
@@ -1,14 +1,24 @@
 <script setup lang="ts">
-import { ref, onMounted } from 'vue';
+import Button from 'primevue/button';
+import { ref, onMounted, Ref } from 'vue';
 import BucketsTable from './BucketsTable.vue';
```
jujaga, line 4: @ reference recommended vs relative.

frontend/src/components/bucket/BucketsTable.vue
```diff
@@ -17,12 +17,16 @@ const permissionsVisible = ref(false);
 const permissionsBucketId = ref('');
 const permissionBucketName = ref('');
 
-const emit = defineEmits(['show-info']);
+const emit = defineEmits(['show-info', 'show-bucket-config']);
```
jujaga, line 20: Alphabetical emit ordering

frontend/src/components/bucket/BucketConfigForm.vue
TimCsaky: not sure we found the best solution by having 'REDACTED' as initial value.
might have to re-think the best UX

frontend/src/components/bucket/BucketConfigDialog.vue
dc-bcgov: div removed.

kamorel: I would refactor this to have this entire <Dialog> directly in the BucketList instead of an extra level of redirection, and delete the BucketConfigDialog.vue file.

frontend/src/components/bucket/BucketList.vue
```diff
@@ -17,6 +27,23 @@ const closeInfo = () => {
...
+    bucketConfigHeader.value = BucketConfigConst.headerNewBucket;
+    bucketConfigTitle.value = bucket.bucketName;
+  } else {
+    bucketConfigHeader.value = BucketConfigConst.headerNewBucket;
+    bucketConfigTitle.value = BucketConfigConst.titleNewBucket;
+  }
+  isBucketConfigUpdate.value = isUpdate;
+  bucketToUpdate.value = bucket;
+  displayBucketConfig.value = true;
+};
```
kamorel, line 41: Can bucketConfigHeader be baked directly into the modal as there is no difference here? Then bucketConfigTitle can be set via ternary operator.

frontend/src/components/bucket/BucketConfigDialog.vue
dc-bcgov: Personally, I am not a big fan of big files/components (100+ lines) - would rather brake down to smaller pieces - easier for testing too. If @jujaga and @TimCsaky agree to refactor in that way, sure no problem. Please bear in mind this will take additional time to test again and etc.

TimCsaky: not too bothered. i like it as is.

frontend/src/components/bucket/BucketList.vue
```diff
@@ -17,6 +27,23 @@ const closeInfo = () => {
   displayInfo.value = null;
 };
 
+const showBucketConfig = (isUpdate: boolean, bucket: Bucket) => {
```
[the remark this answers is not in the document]
dc-bcgov: isUpdate is removed from all three components.

frontend/src/components/bucket/BucketConfigDialog.vue
jujaga: I'm going to agree with @kamorel on this. While it's fine making more smaller components, they should really only be done if these components have some degree of reuse outside of the one parent component calling it. Since this is very focused to Bucket addition flow only, there's no merit splitting this apart as you then have to do a bunch of extra watchers and emit listeners to be able to keep state synced between parent and child.

dc-bcgov: Roger that. Refactoring.

frontend/src/components/bucket/BucketList.vue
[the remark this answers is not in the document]
kamorel: The function definition needs to be changed as well. The isUpdate param is not necessary.
`const showBucketConfig = (bucket?: Bucket) => {`

frontend/src/components/bucket/BucketList.vue
```diff
@@ -17,6 +27,18 @@ const closeInfo = () => {
   displayInfo.value = null;
 };
 
+const showBucketConfig = (bucket?: Bucket) => {
+
```
kamorel, line 31: newline

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -0,0 +1,151 @@
+<script setup lang="ts">
+import Button from 'primevue/button';
+import { Form } from 'vee-validate';
+import { object, string } from 'yup';
+import { useBucketStore } from '@/store';
+import { useToast } from 'primevue/usetoast';
+import Password from '@/components/form/Password.vue';
+import TextInput from '@/components/form/TextInput.vue';
```
kamorel, line 8: alphabetize imports

frontend/src/components/form/Password.vue
```diff
@@ -0,0 +1,39 @@
...
+</script>
+
+<template>
+  <div class="field">
+    <label :for="name">{{ label }}</label>
+    <Password
+      v-model="value"
+      :aria-describedby="`${name}-help`"
+      :name="name"
+      :type="type"
```
TimCsaky, line 23: not a valid attribute. can be removed

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -0,0 +1,151 @@
...
+}
+
+:deep(.pi-eye-slash) {
+  right: auto !important;
+  margin-left: 10px;
+}
+
+button {
+  margin-top: 20px;
+}
```
kamorel, line 150: Can this be replaced with primeflex margins?
https://www.primefaces.org/primeflex/margin

frontend/src/components/bucket/BucketList.vue
```diff
@@ -1,14 +1,24 @@
...
+import Button from 'primevue/button';
+import Dialog from 'primevue/dialog';
+import { ref, onMounted, Ref } from 'vue';
+import BucketConfigForm from '@/components/bucket/BucketConfigForm.vue';
+import { BucketConfig as BucketConfigConst } from '@/utils/constants';
+import BucketsTable from '@/components/bucket/BucketsTable.vue';
 import BucketsSidebar from './BucketsSidebar.vue';
 import { useBucketStore } from '@/store';
 import { useToaster } from '@/composables/useToaster';
+import type { Bucket } from '@/interfaces';
```
kamorel, line 12: Import ordering, and a relative path that should be changed to an absolute.

TimCsaky (approved):
looks good. I suspect we may want to revisit the decision to show 'REDACTED'.

kamorel (approved):

jujaga (approved):

PR #40, opened 2023-02-24
File Details load object

loneil:
# Description
The File Details page is on a route but doesn't get details on mount, just takes in the table row object. So change it to call the API on mount.

All the structures on the page are COMSObject Types so we can just use the same object search from the table, providing a specific singular object ID to search permissions and object and metadata etc.

This works loading data when navigating to the page, though on hosted URLs has the same issue with the params as all other pages, so essentially paves the way to be fixed by the general solution from https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3002

## Checklist


kamorel (approved):

dc-bcgov (approved):

loneil:
> I guess we need to decide when this gets merged with upcoming store refactor.

Don't see any conflict lines (only 1 line changed in store) compared to Jeremy's PR.

jujaga:
> > I guess we need to decide when this gets merged with upcoming store refactor.
> 
> Don't see any conflict lines (only 1 line changed in store) compared to Jeremy's PR.

This will be merged in after #41.

jujaga (approved):

PR #42, opened 2023-03-01
Style tweaks

loneil:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3025
Toggle outline class of upload buttons to match mockup design. (doesn't address **all** button colors, that would be another task if we're doing that)

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2971
Align header styles with designs. Move subheader text outside so default flex centers things. Make a dialog class for global styling to apply to all dialogs of this type. 

image

## Types of changes

## Checklist

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -52,4 +52,7 @@ plugins:
   sass-lint:
     enabled: true
     rules:
-      nesting-depth: 5
+      nesting-depth:
+        - 2
+        -
+          max-depth: 5
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -51,3 +51,5 @@ plugins:
     enabled: true
   sass-lint:
     enabled: true
+    rules:
+      nesting-depth: 5
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -54,5 +54,4 @@ plugins:
     rules:
       nesting-depth:
         - 2
-        -
-          max-depth: 5
+        - max-depth: 5
--- a/frontend/src/assets/main.scss
+++ b/frontend/src/assets/main.scss
@@ -1,6 +1,6 @@
-@import './base.css';
-@import './primevue.scss';
-@import './variables.scss';
+@import "./base.css";
+@import "./primevue.scss";
+@import "./variables.scss";
 
 #app {
   margin: 0 auto;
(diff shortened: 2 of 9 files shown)


.codeclimate.yml
```diff
@@ -51,3 +51,8 @@ plugins:
     enabled: true
   sass-lint:
     enabled: true
+    rules:
+      nesting-depth:
+        - 2
+        -
+          max-depth: 5
```
loneil, line 58: sass-lint is crazy in that it won't let same-line selectors, but also has a max nesting depth of 2, (so why use sass...)?
We need higher specificity especially to avoid `important` spam over primevue defaults sometimes.

And yes this is the documented config syntax https://github.com/sasstools/sass-lint#rules

TimCsaky:
should we get rid of the sub-heading 'Share' (the second one seems a bit much, no?) @tpantella

TimCsaky (approved):

tpantella:
@TimCsaky I agree. In [mockups](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2933?src=confmacro), it isn't jarring because the filename is integrated in the top heading. In this, because the headings have been altered, I would take out the subheading. So yes, please either put the filename in the first heading as designed or, if the deviation is needed for dev reasons, please remove the second "Share" subheading.

frontend/src/components/object/ObjectUpload.vue
```diff
@@ -67,6 +67,11 @@ const onRemoveUploadedFile = async (index: number) => {
 const onRemoveFailedFile = async (index: number) => {
   failedFiles.value.splice(index, 1);
 };
+
+// Based on files prop from upload component, are we in ready to upload mode
+const noFilesChosen = (files: File[]) => {
+  return !files || files.length === 0;
+};
```
jujaga, line 74: TS style - we're aiming to standardize array notations on angle bracket notation. Also the following would be equivalent leveraging null cascading syntax.
```ts
const noFilesChosen = (files?: Array<File>): boolean => !x?.length;
```

jujaga (approved):
Generally looks good. Pre-approving but with a few minor comments.

.codeclimate.yml
loneil: It should work that way, though I saw the sass-lint documentation and issues do it as above in multiple spots so wasn't sure if it's more or less confusing to match that documentation or not

frontend/src/assets/main.scss
```diff
@@ -27,3 +28,31 @@
 .permissions-modal {
   width: 800px;
 }
+
+// Info/settings dialog modals
+.bcbox-info-dlg {
```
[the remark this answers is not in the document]
loneil: I'll use 'dialog'

frontend/src/components/object/ObjectUpload.vue
loneil: fixed

frontend/src/assets/variables.scss
```diff
@@ -1 +1,2 @@
-// Add as needed
+// General font/etc colors
+$bcbox-icon-primary: #036;
```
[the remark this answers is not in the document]
jujaga: That seems like a silly CC rule but ok we'll run with that for now then. We can re-evaluate later when we do the larger consistency pass.

jujaga (approved):

kamorel (approved):

PR #46, opened 2023-03-09
Configurable banner

loneil:
# Description

Add a configurable notification banner. Could be expanded to support multiple independent "operations notifications" in different spots if that were ever needed. Set helm values for envs.

Won't see these in PR deployment til helm value update goes to dev env.

AC from https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3070?

image

image

## Checklist


app/config/default.json
```diff
@@ -1,6 +1,10 @@
 {
   "frontend": {
-    "apiPath": "api/v1"
+    "apiPath": "api/v1",
+    "oidc": {
+      "authority": "https://dev.loginproxy.gov.bc.ca/auth/realms/standard",
+      "clientId": "bc-box-local-4545"
+    }
```
jujaga, line 7: Please revert this from history. There is no business having these default configurables appear in the codebase.

.github/environments/values.dev.yaml
```diff
@@ -6,6 +6,7 @@ config:
     FRONTEND_COMS_APIPATH: https://coms-dev-master.apps.silver.devops.gov.bc.ca/api/v1
     FRONTEND_OIDC_AUTHORITY: https://dev.loginproxy.gov.bc.ca/auth/realms/standard
     FRONTEND_OIDC_CLIENTID: bc-box-4555
+    FRONTEND_NOTIFICATION_HOME_BANNER: This is the DEV environment of BCBox. Uploaded files may not persist and may be deleted from the COMS database as we continue development. Please do not upload personal or private information. Thank you for your understanding as we work to improve BCBox
```
jujaga, line 9: Please have this be inserted in alphabetical order. In this case, it should come before the OIDC keys.

app/config/custom-environment-variables.json
```diff
@@ -7,6 +7,9 @@
     "oidc": {
       "authority": "FRONTEND_OIDC_AUTHORITY",
       "clientId": "FRONTEND_OIDC_CLIENTID"
+    },
+    "notifications": {
+      "homeBanner": "FRONTEND_NOTIFICATION_HOME_BANNER"
```
jujaga, line 12: Please have this be inserted in alphabetical order. In this case, it should come before the OIDC keys.

frontend/src/components/bucket/BucketList.vue
```diff
@@ -50,6 +58,12 @@ onMounted(async () => {
     <div>
       <h1>Select a bucket</h1>
       <h3>Buckets are containers for storing objects.</h3>
+      <Message
+        v-if="getConfig?.notifications?.homeBanner"
+        severity="warn"
+      >
+        {{ getConfig?.notifications?.homeBanner }}
+      </Message>
```
jujaga, line 66: Do we want to be mounting here, or would it be a better idea to make this feature more generic and mounted higher up the views/router tree?

charts/bcbox/values.yaml
```diff
@@ -115,6 +115,7 @@ config:
     FRONTEND_COMS_APIPATH: ~
     FRONTEND_OIDC_AUTHORITY: ~
     FRONTEND_OIDC_CLIENTID: ~
+    FRONTEND_NOTIFICATION_HOME_BANNER: ~
```
jujaga, line 118: Please have this be inserted in alphabetical order. In this case, it should come before the OIDC keys.
Also note that since we are doing a helm chart change, we need to increment the patch version of this chart in Charts.yaml for this to take any effect after it gets merged.

.github/environments/values.pr.yaml
```diff
@@ -1 +1,5 @@
 ---
+config:
+  enabled: true
+  configMap:
+    FRONTEND_NOTIFICATION_HOME_BANNER: This is a PR environment of BCBox. Uploaded files may not persist and may be deleted from the COMS database as we continue development. Please do not upload personal or private information. Thank you for your understanding as we work to improve BCBox
```
loneil, line 5: Trying out, not sure if a single override is kosher here

.github/environments/values.dev.yaml
loneil: Fixed

app/config/custom-environment-variables.json
loneil: Fixed

charts/bcbox/values.yaml
loneil: Fixed, bumped chart version

frontend/src/components/bucket/BucketList.vue
loneil: Yeah I think there could be a "top level" notification that appears above the header in the global layout. But this one was for the spots defined by Sharolyn in the ticket. We could add other notifications in this context as well.

dc-bcgov (changes requested):

frontend/src/components/bucket/BucketList.vue
```diff
@@ -2,9 +2,13 @@
 import { storeToRefs } from 'pinia';
 import { ref, onMounted } from 'vue';
 
-import { BucketConfigForm, BucketSidebar, BucketTable } from '@/components/bucket';
-import { Button, Dialog } from '@/lib/primevue';
-import { useAuthStore, useBucketStore } from '@/store';
+import {
```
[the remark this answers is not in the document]
loneil: Not sure? I didn't change this (just formatting adjustment). It seems to be the same in the other components I see

charts/bcbox/values.yaml
```diff
@@ -113,6 +113,7 @@ config:
   configMap:
     FRONTEND_APIPATH: "api/v1"
     FRONTEND_COMS_APIPATH: ~
+    FRONTEND_NOTIFICATION_HOME_BANNER: ~
```
TimCsaky, line 116: not to be pedantic, but it might make sense to label this `FRONTEND_NOTIFICATION_BANNER`  and with property `.banner` in the env and components as it doesnt only appear on homepage (?)

TimCsaky (approved):

frontend/src/components/bucket/BucketList.vue
jujaga: I could see us inserting this into `bcbox/AppLayout.vue` between the header and main views slots...
```vue
    <!-- Header/Nav -->
    <div class="layout-head">
      <Header />
      <slot name="nav" />
    </div>

    <!-- Insert the "Global Notification Component" here -->

    <!-- Main views -->
    <main class="layout-container">
      <slot name="main" />
    </main>
```

charts/bcbox/values.yaml
jujaga: Yea after thinking about this some more, it might make more sense for the ENV variable name to just be `FRONTEND_NOTIFICATION_BANNER` as Tim recommended. In the json object, we'd want it to then just be `frontend.notificationBanner` or something similar. It'd be less verbose and would be able to cover more aspects and situations.

loneil: Changing to
image

frontend/src/components/bucket/BucketList.vue
jujaga: To be considered at a later date 👍

jujaga (approved):

PR #51, opened 2023-03-15
Fix/3098 small bugs in test

dc-bcgov:
# Description

## Checklist

### DIFF
--- a/frontend/src/components/bucket/BucketSidebar.vue
+++ b/frontend/src/components/bucket/BucketSidebar.vue
@@ -1,10 +1,10 @@
 <script setup lang="ts">
 import { Button } from '@/lib/primevue';
-import { storeToRefs } from 'pinia';
 import { onMounted, ref, Ref } from 'vue';
+import { storeToRefs } from 'pinia';
 
-import { usePermissionStore, useUserStore } from '@/store';
 import { Permissions } from '@/utils/constants';
+import { usePermissionStore, useUserStore } from '@/store';
 
 import type { Bucket } from '@/types';
 import type { BucketPermission } from '@/types';
--- a/frontend/src/components/bucket/BucketSidebar.vue
+++ b/frontend/src/components/bucket/BucketSidebar.vue
@@ -1,7 +1,13 @@
 <script setup lang="ts">
 import { Button } from '@/lib/primevue';
+import { storeToRefs } from 'pinia';
+import { onMounted, ref, Ref } from 'vue';
+
+import { usePermissionStore, useUserStore } from '@/store';
+import { Permissions } from '@/utils/constants';
 
 import type { Bucket } from '@/types';
+import type { BucketPermission } from '@/types';
 
 // Props
 type Props = {
@@ -10,6 +16,16 @@ type Props = {
 
 const props = withDefaults(defineProps<Props>(), {});
 
+//Store
+const permissionStore = usePermissionStore();
+const userStore = useUserStore();
+
+const { getBucketPermissions } = storeToRefs(permissionStore);
+const { userSearch } = storeToRefs(userStore);
+
+// State
+const managedBy: Ref<string | undefined> = ref();
+
 // Emits
 const emit = defineEmits(['close-sidebar-info']);
 
@@ -17,6 +33,24 @@ const emit = defineEmits(['close-sidebar-info']);
 const closeSidebarInfo = async () => {
   emit('close-sidebar-info');
 };
+
+// Actions
+async function load() {
+  await permissionStore.fetchBucketPermissions({bucketId: props.sidebarInfo.bucketId});
+
+  const uniqueIds = [...new Set(getBucketPermissions.value
+    .filter( (x: BucketPermission) => x.bucketId === props.sidebarInfo.bucketId && x.permCode === Permissions.MANAGE )
+    .map( (x: BucketPermission) => x.userId) )];
+
+  if(uniqueIds.length) {
+    await userStore.fetchUsers({userId: uniqueIds} );
+    managedBy.value = userSearch.value.map( x => x.fullName ).join( ', ');
+  }
+}
+
+onMounted(() => {
+  load();
+});
 </script>
 
 <template>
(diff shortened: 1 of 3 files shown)


frontend/src/components/bucket/BucketSidebar.vue
```diff
@@ -1,7 +1,13 @@
 <script setup lang="ts">
 import { Button } from '@/lib/primevue';
```
kamorel, line 2: Import ordering. This is an application import.

frontend/src/components/bucket/BucketSidebar.vue
```diff
@@ -1,7 +1,13 @@
...
+import { onMounted, ref, Ref } from 'vue';
+import { storeToRefs } from 'pinia';
+
+import { Permissions } from '@/utils/constants';
+import { usePermissionStore, useUserStore } from '@/store';
 
 import type { Bucket } from '@/types';
+import type { BucketPermission } from '@/types';
```
kamorel, line 10: Can be one line. Same import source.

frontend/src/components/bucket/BucketSidebar.vue
```diff
@@ -10,13 +16,41 @@ type Props = {
 
 const props = withDefaults(defineProps<Props>(), {});
 
+//Store
```
kamorel, line 19: Missing spacing after //

frontend/src/components/bucket/BucketSidebar.vue
```diff
@@ -10,13 +16,41 @@ type Props = {
...
+
 // Emits
 const emit = defineEmits(['close-sidebar-info']);
 
 // Actions
 const closeSidebarInfo = async () => {
   emit('close-sidebar-info');
 };
+
+// Actions
```
kamorel, line 37: Duplicate comment line

frontend/src/components/bucket/BucketSidebar.vue
```diff
@@ -10,13 +16,41 @@ type Props = {
...
+const userStore = useUserStore();
+
+const { getBucketPermissions } = storeToRefs(permissionStore);
+const { userSearch } = storeToRefs(userStore);
+
+// State
+const managedBy: Ref<string | undefined> = ref();
+
 // Emits
 const emit = defineEmits(['close-sidebar-info']);
```
kamorel, line 30: Please move the Emits to after the Props. This is not outlined anywhere, but a pattern I am moving to for consistency is Props/Emits/Store/State/Actions.

frontend/src/components/bucket/BucketSidebar.vue
```diff
@@ -10,13 +16,41 @@ type Props = {
...
 // Actions
 const closeSidebarInfo = async () => {
   emit('close-sidebar-info');
 };
+
+// Actions
+async function load() {
+  await permissionStore.fetchBucketPermissions({bucketId: props.sidebarInfo.bucketId});
+
+  const uniqueIds = [...new Set(getBucketPermissions.value
```
TimCsaky, line 39: when testing this PR in browser, I have 2 buckets: bucket1 only has me as manager. and nr-iit-showcase bucket.
if i open sidebar on bucket1 it shows just me as manager. if i then open sidebar on nr-iit-showcase bucket it again shows only me as manager. (if i refresh page  nr-iit-showcase shows all managers,)
`getBucketPermissions.value` is not pulling the latest state.
i wonder if it's because we created   `getBucketPermissions` const at the top of this file on line 23.
either that or we have a timing issue.

TimCsaky (commented):
you need to resolve a conflict

frontend/src/components/bucket/BucketSidebar.vue
```diff
@@ -13,10 +18,37 @@ const props = withDefaults(defineProps<Props>(), {});
...
+
+  if(uniqueIds.length) {
+    await userStore.fetchUsers({userId: uniqueIds} );
+    managedBy.value = userSearch.value.map( x => x.fullName ).join( ', ');
+  }
+}
+
+onMounted(() => {
+  load();
+});
```
jujaga, line 51: We should strongly consider using an earlier Vue lifecycle hook for this. I would recommend we use the `onBeforeMounted` event hook instead of `onMounted`. As well, since load is only ever used in one place, that can be consolidated to be the body of the `onBeforeMounted` callback function instead.

```ts
onBeforeMounted(() => {
  await permissionStore.fetchBucketPermissions({bucketId: props.sidebarInfo.bucketId});
  const uniqueIds = [...new Set(getBucketPermissions.value
    .filter( (x: BucketPermission) => x.bucketId === props.sidebarInfo.bucketId && x.permCode === Permissions.MANAGE )
    .map( (x: BucketPermission) => x.userId) )];
  if(uniqueIds.length) {
    await userStore.fetchUsers({userId: uniqueIds} );
    managedBy.value = userSearch.value.map( x => x.fullName ).join(', ');
  }
});

**Review by jujaga — COMMENTED** [review]
Double check import ordering. While the groupings are correct, you also want to do a secondary sort based on the from string value.

File: frontend/src/components/bucket/BucketSidebar.vue [hunk 8]
```diff
@@ -1,7 +1,12 @@
 <script setup lang="ts">
+import { storeToRefs } from 'pinia';
+import { onBeforeMount, ref, Ref, watch } from 'vue';
+
+import { Permissions } from '@/utils/constants';
 import { Button } from '@/lib/primevue';
+import { usePermissionStore, useUserStore } from '@/store';
```
kamorel, line 7: @/lib
@/store
@/utils

frontend/src/components/bucket/BucketSidebar.vue
```diff
@@ -1,7 +1,12 @@
 <script setup lang="ts">
+import { storeToRefs } from 'pinia';
+import { onBeforeMount, ref, Ref, watch } from 'vue';
```
kamorel, line 3: Ref is actually a type
`import type { Ref } from 'vue';`

kamorel (approved):

TimCsaky (approved):
all working for me now

jujaga (approved):

PR #52, opened 2023-03-15
Restructure network calls for Object sidebar & File Details

kamorel:
# Description

1. Bump PrimeVue to 3.25.0
2. Clean up unnecessary network calls
3. Various bug fixes

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/form/SearchUsers.vue
+++ b/frontend/src/components/form/SearchUsers.vue
@@ -6,6 +6,7 @@ import { Button, Dropdown, RadioButton } from '@/lib/primevue';
 import { useConfigStore, useUserStore } from '@/store';
 import { Regex } from '@/utils/constants';
 
+import type { IChangeEvent, IInputEvent } from '@/interfaces';
 import type { IdentityProvider, User, UserPermissions } from '@/types';
 
 // Props
@@ -24,9 +25,9 @@ const { getConfig } = storeToRefs(useConfigStore());
 const { userSearch } = storeToRefs(useUserStore());
 
 // State
+const invalidSelectedUser: Ref<boolean> = ref(false);
 const selectedIDP: Ref<IdentityProvider | null> = ref(null);
 const selectedUser: Ref<User | null> = ref(null);
-const selectedUserIsInvalid: Ref<boolean> = ref(false);
 const userSearchInput: Ref<string | undefined> = ref('');
 const userSearchPlaceholder: Ref<string | undefined> = ref('');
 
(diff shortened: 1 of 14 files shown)


jujaga (approved):
Semantically makes sense. Assuming the library bump doesn't break anything 👍

frontend/src/components/object/ObjectMetadata.vue
```diff
@@ -23,9 +23,6 @@ const objectMetadata: Ref<Metadata | undefined> = ref(undefined);
 
 // Actions
 async function load() {
-  if( props.fullView ) {
```
loneil, line 26: Is fullView unused now and removeable?

loneil (dismissed):
Wierdest thing.... on this one (pr deploy or localhost) I can't type in the user search fields any more. Blocking til I can confirm

image

kamorel:
> Wierdest thing.... on this one (pr deploy or localhost) I can't type in the user search fields any more. Blocking til I can confirm
> 
> image

Should be resolved now.

frontend/src/components/form/SearchUsers.vue
```diff
@@ -53,18 +53,17 @@ const onCancel = () => {
...
-    selectedUser.value = event.value;
-  }
-  else {
+      selectedUser.value = event.value;
+    }
+    else {
     // Invalid
-    selectedUserIsInvalid.value = true;
+      selectedUserIsInvalid.value = true;
+    }
```
jujaga, line 65: We should be careful about how we're naming our variables especially if they have the potential to be negative affirmations. In this case, I can see why it was named invalid (p-invalid class), but I would consider renaming this to something more assertive like `invalidSelectedUser` or `isInvalidSelectedUser` - by declaring invalid earlier it makes it more salient that we're focusing on the negative case instead of emphasizing on the user.

frontend/src/components/form/SearchUsers.vue
```diff
@@ -74,11 +73,17 @@ const onInput = async (event: any) => {
       if( input.length >= 3 ) {
         await userStore.fetchUsers({ idp: selectedIDP.value.idp, search: input });
       }
+      else {
+        userStore.clearSearch();
+      }
```
jujaga, line 78: 1) Is there a way we can determine what kind of type we are getting for the event callback? We'd want to avoid using the any type wherever possible. If no such type exists in the vue types, then we should be more specific with the declaration.

2) While the nested boolean logic here makes sense due to necessity, it is getting dangerously close to a high cognitive complexity level. I propose we do the following to attempt to flatten out its structure:

```ts
const onInput = async (event: { target: { value: string } }) => {
  const input: string = event.target.value;
  if ( selectedIDP.value?.idp ) {
    if ( selectedIDP.value?.searchable && input.length >= 3 ) {
      await userStore.fetchUsers({ idp: selectedIDP.value.idp, search: input });
    } else if ( input.match( Regex.EMAIL ) ) {
      await userStore.fetchUsers({ idp: selectedIDP.value.idp, email: input });
    } else {
      userStore.clearSearch();
    }
  }
};
```
3) Even with the above, do we actually need the async/await on this event? We're not explicitly needing to do an operation right after that requires the blocking aspect of the await, so we might be able to get away with making this a synchronous callback function instead, and let the asynchronous userStore actions complete at their leisure.

jujaga (approved):

dc-bcgov (approved):

PR #53, opened 2023-03-16
disable other stuff when uploadin

loneil:
# Description

_Still need to test out a bit more but should get some feedback on implementation_

When uploading files, disable other parts of the object list page
image

"v2" mocks from here https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-2913?

Use local field to track "file upload open" bool. Pass to any children if they are to be disabled too. If more and more components are added and they're children the disabled prop will have to keep being added if they're needing to be disabled as well.
If non-children will need to be disabled might need to to store implementation. Tried that out, it's in commit history of this PR.

Disable buttons. No global disable for datatable so use the usual no-click, opacity css trick.

## Checklist

### DIFF
--- a/frontend/src/components/object/DeleteObjectButton.vue
+++ b/frontend/src/components/object/DeleteObjectButton.vue
@@ -8,7 +8,7 @@ import { ButtonMode } from '@/utils/enums';
 
 // Props
 type Props = {
-  disabled: boolean;
+  disabled?: boolean;
   ids: Array<string>;
   mode: ButtonMode;
 };
--- a/frontend/src/components/object/DeleteObjectButton.vue
+++ b/frontend/src/components/object/DeleteObjectButton.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { storeToRefs } from 'pinia';
 import { ref } from 'vue';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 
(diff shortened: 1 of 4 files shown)


kamorel:
isUploadOpen feels more like an application state, let's move this to the app store?
Do we want the upload pane to remain active when navigating away then coming back? Might want to set this to false in OnUnmounted in ObjectList.

kamorel (commented):
isUploadOpen feels more like an application state, let's move this to the app store?
Do we want the upload pane to remain active when navigating away then coming back? Might want to set this to false in OnUnmounted in ObjectList.

jujaga:
> @kamorel yeah I can move it to that store. I **did** set it to false in onMounted there but somehow lost that change... adding it back.

I'm not absolutely convinced `isUploadOpen` should be moved to the appStore because this would only apply in the application context where you would be wanting to upload a file. In this case, it would only ever show up under the `/list/objects` route, so at least for me, it makes sense that this state would reside more on either the associated component or view (I'm slightly leaning towards view as I'd like to eventually change it so that when that dialog shows up, it also invokes an address bar change)

kamorel:
> > @kamorel yeah I can move it to that store. I **did** set it to false in onMounted there but somehow lost that change... adding it back.
> 
> I'm not absolutely convinced `isUploadOpen` should be moved to the appStore because this would only apply in the application context where you would be wanting to upload a file. In this case, it would only ever show up under the `/list/objects` route, so at least for me, it makes sense that this state would reside more on either the associated component or view (I'm slightly leaning towards view as I'd like to eventually change it so that when that dialog shows up, it also invokes an address bar change)

Another option would be to keep it in the local state and add an additional prop to DeleteObjectButton and DownloadObjectButton for disabled

loneil:
I'd used the local var and passed down at first, but then looking at the component and thinking about what else could go in that page in the future I saw some refactoring possible down the line. Including maybe needing to block things that aren't direct children, so did the store state to accommodate that.

Could work as local with props to children at this time though yeah. I think the state would be good for flexibility but if we'd rather have local I can go back to that. The scope and lifecycle are a little cleaner in that case.

kamorel (approved):

jujaga (approved):

PR #55, opened 2023-03-20
Various bug fixes

kamorel:
# Description

1. Prevent removal of the final MANAGE permission. Either through permission removal, or full user removal.
2. Route user to the object list after deleting a file in object details view
3. Reset user selection on any input change in user search

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/object/ObjectPermission.vue
+++ b/frontend/src/components/object/ObjectPermission.vue
@@ -4,11 +4,13 @@ import { onMounted, ref } from 'vue';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 
 import ObjectPermissionAddUser from '@/components/object/ObjectPermissionAddUser.vue';
+import { useAlert } from '@/composables/useAlert';
 import { Button, Checkbox, Column, DataTable } from '@/lib/primevue';
 import { usePermissionStore } from '@/store';
 import { Permissions } from '@/utils/constants';
 
 import type { Ref } from 'vue';
+import type { UserPermissions } from '@/types';
 
 // Props
 type Props = {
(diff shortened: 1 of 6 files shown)


loneil (approved):
Think this checks out, tested out the cases from the ticket and some regression.
Would suggest the following at some point too
- success Toast when file is deleted (either from list, or file details page)
- Consider loading state when removing a user from the permissions list, hangs _just_ perceptively long enough

frontend/src/components/object/ObjectPermission.vue
```diff
@@ -25,19 +27,40 @@ const { getMappedObjectToUserPermissions } = storeToRefs(permissionStore);
...
+      removeManageAlert.show();
+
+      // Set the value back as clicking will automatically change it
+      const perm: UserPermissions = getMappedObjectToUserPermissions.value
+        .find( (x: UserPermissions) => x.userId === userId ) as UserPermissions;
+      perm.manage = true;
+    }
+    else {
+      permissionStore.deleteObjectPermission(props.objectId, userId, permCode);
+    }
```
jujaga, line 63: I can't quite tell from just visual inspection, but does this logic also factor in situations where a user may have manage permission on a bucket level, but for the object they're looking at, there's no explicit permissions bound to it? May need to think further on edge cases for this, but it may be permssible to allow an object to have no users bound to it at all, as users will have a way to escape hatch it and reapply perms if needed by virtue of having bucket level permissions cascading downwards?

TimCsaky: I believe we should allow someone with MANAGE perm (either at obj or bucket level) to remove whoever they want as long as at least one person is always left who has the MANAGE perm.

To get to object permissions modal, the current user must have MANAGE perm at either bucket-level or object-level.
I think we only need to prevent the current user from de-selecting MANAGE (or removing user) for **themself** and they are the last user in the object permissions list with object-level MANAGE perm.

dc-bcgov (approved):

jujaga (approved):

PR #56, opened 2023-03-20
Wrap object and bucket name heading

TimCsaky:
# Description

- fixes layout issues where object or bucket name overruns the screen and doesnt wrap to two lines.
- datatables have font-size 16px, set to 15px to match custom default set in body.
- I found some other css/display issues that we'll need to get fixed separately

## Checklist

### DIFF
--- a/frontend/src/assets/base.css
+++ b/frontend/src/assets/base.css
@@ -66,12 +66,7 @@ body {
   line-height: 1.6;
   font-family: Inter, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Fira Sans', 'Droid Sans', 'Helvetica Neue',
     sans-serif;
-  font-size: 15px;
   text-rendering: optimizeLegibility;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
-}
-
-.p-datatable{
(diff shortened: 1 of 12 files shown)


frontend/src/assets/base.css
```diff
@@ -71,3 +71,7 @@ body {
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
 }
+
+.p-datatable{
+  font-size: 15px;
+}
```
jujaga, line 77: Not sure if we want to be doing this. This feels a bit like a magic number.

loneil: I think it's fine (would use `rem` setting over px though since it's the main text on the page) if the UX decision is to have the table font be specifically different. Otherwise I would say it should be the main body font size since it's kind of the main content of the app?

dc-bcgov (changes requested):
Just fix two things we talked about, and it looks good to me.

loneil:
Can probably address the code climate suggestions, wonder if linter/formatter doesn't do the scss files in the same way CC expects... maybe something to look at later

dc-bcgov (approved):

jujaga (approved):

kamorel:
A number of linting items that should be looked at before merging.

kamorel (approved):

PR #57, opened 2023-03-21
Provide clarity on the Create Bucket screen

dc-bcgov:
Modified configure bucket form to support new labels/placeholders/helptext/mandatory described in https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3103. Changed custom common 'InputText' and 'Password' vue components to support helpText rendering. Removed 'Key' from validation schema, it's not mandatory anymore.

Screenshot 2023-03-21 at 1 23 25 PM

# Description

## Checklist

### DIFF
--- a/frontend/src/components/bucket/BucketConfigForm.vue
+++ b/frontend/src/components/bucket/BucketConfigForm.vue
@@ -43,7 +43,7 @@ const schema = object({
   endpoint: string().max(255).required().label('Endpoint'),
   accessKeyId: string().max(255).required().label('Access Key ID'),
   secretAccessKey: string().max(255).required().label('Secret Access Key'),
-  key: string().max(255).label('Key'),
+  key: string().max(255).label('Key')
 });
 
 // Actions
--- a/frontend/src/components/bucket/BucketConfigForm.vue
+++ b/frontend/src/components/bucket/BucketConfigForm.vue
@@ -58,7 +58,7 @@ const onSubmit = async (values: any) => {
       accessKeyId: values.accessKeyId !== 'REDACTED' ? values.accessKeyId : undefined,
       secretAccessKey:
         values.secretAccessKey !== 'REDACTED' ? values.secretAccessKey : undefined,
-      key: values.key.trim() ? values.key : '/',
+      key: values.key ? values.key : '/',
       active: true
     } as Bucket;
 
--- a/frontend/src/components/bucket/BucketConfigForm.vue
+++ b/frontend/src/components/bucket/BucketConfigForm.vue
@@ -58,7 +58,7 @@ const onSubmit = async (values: any) => {
       accessKeyId: values.accessKeyId !== 'REDACTED' ? values.accessKeyId : undefined,
       secretAccessKey:
         values.secretAccessKey !== 'REDACTED' ? values.secretAccessKey : undefined,
-      key: values.key,
+      key: values.key.trim() ? values.key : '/',
       active: true
     } as Bucket;
 
--- a/frontend/src/components/bucket/BucketConfigForm.vue
+++ b/frontend/src/components/bucket/BucketConfigForm.vue
@@ -43,7 +43,7 @@ const schema = object({
   endpoint: string().max(255).required().label('Endpoint'),
   accessKeyId: string().max(255).required().label('Access Key ID'),
   secretAccessKey: string().max(255).required().label('Secret Access Key'),
-  key: string().max(255).required().label('Key'),
+  key: string().max(255).label('Key'),
 });
 
 // Actions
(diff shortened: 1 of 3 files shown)


frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -104,27 +104,39 @@ const onCancel = () => {
...
+        label="Endpoint *"
+        placeholder="https://exampleendpoint.gov.bc.ca/"
+        helptext="The URL that specifies how to access the bucket."
       />
       <Password
         name="accessKeyId"
-        label="Access key Identifier"
+        label="Access key identifier / User ID *"
+        placeholder="AKIAIOSFODNN7EXAMPLE"
+        helptext="A public identifier used to access the bucket. Also known as User ID."
```
TimCsaky, line 127: this description is not accurate.
maybe it could be : `Unique user/account identifier or username`

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -104,27 +104,39 @@ const onCancel = () => {
...
-        label="Endpoint"
+        label="Endpoint *"
+        placeholder="https://exampleendpoint.gov.bc.ca/"
+        helptext="The URL that specifies how to access the bucket."
       />
       <Password
         name="accessKeyId"
-        label="Access key Identifier"
+        label="Access key identifier / User ID *"
+        placeholder="AKIAIOSFODNN7EXAMPLE"
```
TimCsaky, line 126: you could set the placeholder to be `eg: 'nr-climate-user1'`

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -104,27 +104,39 @@ const onCancel = () => {
...
+        label="Bucket *"
+        placeholder="bucket0123456789"
+        helptext="A unique identifier for a bucket."
       />
       <TextInput
         name="endpoint"
-        label="Endpoint"
+        label="Endpoint *"
+        placeholder="https://exampleendpoint.gov.bc.ca/"
+        helptext="The URL that specifies how to access the bucket."
```
TimCsaky, line 121: i would suggest `The URL of the object storage server`

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -104,27 +104,39 @@ const onCancel = () => {
     >
       <TextInput
         name="bucketName"
-        label="Bucket name (what other users will see)"
+        label="Bucket name *"
+        placeholder="nr-costcontainment-dev"
```
TimCsaky, line 108: maybe the bucket name example should be more user-friendly to distinguish it from the examples of other fields. eg: `eg: 'My Tax Documents'`

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -104,27 +104,39 @@ const onCancel = () => {
...
+        label="Access key identifier / User ID *"
+        placeholder="AKIAIOSFODNN7EXAMPLE"
+        helptext="A public identifier used to access the bucket. Also known as User ID."
       />
       <Password
         name="secretAccessKey"
-        label="Secret access key"
+        label="Secret access key *"
+        placeholder="wjairXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY"
+        helptext="A private key used to authenticate and authorize access to a bucket."
```
TimCsaky, line 133: maybe this should be `A password used to authorize access to a bucket`

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -104,27 +104,39 @@ const onCancel = () => {
...
       />
       <TextInput
         name="key"
         label="Key"
+        placeholder="folder/"
```
TimCsaky, line 138: I dont think it needs the final slash. maybethe placeholder should be `eg: 'documents/reports'`

TimCsaky (commented):
image
Tyler and I think the error message should be red and same size as the description and ideally indented to align with the input text. would be good to see what that looks like if possible

frontend/src/components/bucket/BucketConfigForm.vue
jujaga: I like that recommendation. 'My Important Documents' might be a good one to have (less generic than tax but still good example). Want to make sure there's a mention that this must be a name that hasn't been taken by others yet in this COMS instance.

frontend/src/components/bucket/BucketConfigForm.vue
jujaga: Can do `https://*.objectstore.gov.bc.ca` as a reasonable example. Still forces users to have to change it up, but we can "reasonably" expect most urls for now to match that shape for placeholder.

frontend/src/components/bucket/BucketConfigForm.vue
jujaga: This doesn't even need to be unique, so `User/Account identifier or username` suffices.

frontend/src/components/bucket/BucketConfigForm.vue
jujaga: That or `Password used to access the bucket`

frontend/src/components/bucket/BucketConfigForm.vue
jujaga: As mentioned previously, we need to decide what our general default recommendation behavior should be. it may suffice to just have it be empty and we manipulate the value in service layer in the event the string is empty or undefined, OR we placeholder this to make sure users put in at least a slash or equivalent. Unsure of best course of action.

frontend/src/components/form/Password.vue
```diff
@@ -5,13 +5,15 @@ import { useField, ErrorMessage } from 'vee-validate';
 import { Password } from '@/lib/primevue';
 
 type Props = {
+  helptext?: string;
```
jujaga, line 8: Let's try to alphabetize the props.

jujaga (approved):

kamorel (approved):

PR #61, opened 2023-03-27
More UI qa feedback fixes

loneil:
# Description

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3158
Show short form UUID in table with hover for full one

image

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3153
Disable for public until unauthed File Details page works

image

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/utils/formatters.ts
+++ b/frontend/src/utils/formatters.ts
@@ -37,3 +37,13 @@ export function toKebabCase(str: string | null) {
   const strs = str && str.match(/[A-Z]{2,}(?=[A-Z][a-z]+[0-9]*|\b)|[A-Z]?[a-z]+[0-9]*|[A-Z]|[0-9]+/g);
   return strs ? strs.join('-').toLocaleLowerCase() : '';
 }
+
+/**
+ * @function formatShortUuid
+ * Converts a UUID identifier to just show the first 8 characters 
+ * @param {String} value A typical UUID used from COMS
+ * @returns {String} The first 8 chars
+ */
+export function formatShortUuid(uuid: string) {
+  return uuid ? uuid.slice(0,8) : uuid;
+}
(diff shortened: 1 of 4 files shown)


frontend/src/utils/formatters.ts
```diff
@@ -37,3 +37,13 @@ export function toKebabCase(str: string | null) {
...
 }
+
+/**
+ * @function formatShortUuid
+ * Converts a UUID identifier to just show the first 8 characters 
+ * @param {String} value A typical UUID used from COMS
+ * @returns {String} The first 8 chars
+ */
+export function formatShortUuid(uuid: string) {
+  return uuid ? uuid.slice(0,8) : uuid;
```
dc-bcgov, line 48: Wondering, as this is a common function now, can we have "The first 8 chars" to be parametrized?

loneil: If that's the case, could just use `slice(0,8)` anywhere if the caller needs to specify length. I'd had the point of this one being to standardize the format anywhere some one calls it from (so then if that standard format changes the 8 would only be needed to be changed in one spot rather than from each caller).

If we want to have the param be callable (if there's reasons for variable format), I think probably just remove this function then

jujaga: Slice is a simple enough Array.prototype function I don't think we should be making a util function just for this, as it's part of the base language.

loneil: Yeah util function was essentially a combo of the 8 constant and the way to do it. It's only used once at this point so can remove to the component.

frontend/tests/unit/utils/formatters.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { toKebabCase, formatShortUuid } from '@/utils/formatters';
+
+describe.skip('formatters.ts toKebabCase', () => {
```
jujaga, line 3: Were these tests intended to be skipped?

frontend/tests/unit/utils/formatters.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+    expect(toKebabCase('abc')).toEqual('abc');
+  });
+
+  it('returns blanks if blank provided', () => {
+    expect(toKebabCase('')).toEqual('');
+    expect(toKebabCase(null)).toEqual(null);
+  });
+});
+
+describe.skip('formatters.ts formatShortUuid', () => {
```
jujaga, line 16: Were these tests intended to be skipped?

kamorel (approved):

jujaga (approved):

dc-bcgov (approved):

PR #62, opened 2023-03-27
Proposal: File search, object refresh

loneil:
# Description

Implementation of (in-memory, primevue table) file search. Use primevue filter from https://primevue.org/datatable/#filter
(note there's a current Typescript bug that they keep trying to fix without success, see comment in code)

case-insensitive 'contains' string search using name column only. This could be expanded to other columns if desired
image

Handles the truncation alright as well
image

Not sure if this is the exact UX/placement for this, could adjust but looks fine to me. Maybe check with Tyler.
Could remove darker background and/or borders

Also add object list refresh button while we're at it next to search.
 

## Checklist

### DIFF
--- a/frontend/src/components/object/ObjectTable.vue
+++ b/frontend/src/components/object/ObjectTable.vue
@@ -1,6 +1,5 @@
 <script setup lang="ts">
 import { storeToRefs } from 'pinia';
-import { FilterMatchMode } from 'primevue/api';
 import { ref, watch } from 'vue';
 
 import {
@@ -9,7 +8,7 @@ import {
   ObjectPermission,
   ShareObjectButton
 } from '@/components/object';
-import { Button, Column, DataTable, Dialog, InputText, InputSwitch } from '@/lib/primevue';
+import { Button, Column, DataTable, Dialog, FilterMatchMode, InputText, InputSwitch } from '@/lib/primevue';
 import { useAuthStore, useAppStore, useMetadataStore, useObjectStore, usePermissionStore } from '@/store';
 import { Permissions } from '@/utils/constants';
 import { ButtonMode } from '@/utils/enums';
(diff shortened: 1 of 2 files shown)


frontend/src/components/object/ObjectTable.vue
```diff
@@ -81,12 +82,20 @@ watch( getObjects, async () => {
 watch( selectedObjects, () => {
   objectStore.setSelectedObjects(selectedObjects.value);
 });
+
+// Datatable filter(s)
+const filters = ref({
+  // Need this till PrimeVue gets it together to un-break this again
+  // @ts-ignore
```
loneil, line 89: This FilterMatchMode constant was fine up til 3.24, then it broke, then they had a "fix" that didn't fix it in 3.25, then they had another non-fix in 3.26, and that is still pending (maybe get it right in 3.27 🤷 )
https://github.com/primefaces/primevue/issues/3751#issuecomment-1482963050

Just ignore the 1 line of typescript for now

jujaga: Put in a note to revisit this again whenever we do a primevue update

frontend/src/components/object/ObjectTable.vue
```diff
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import { storeToRefs } from 'pinia';
+import { FilterMatchMode } from 'primevue/api';
```
dc-bcgov, line 3: Couldn't find too much documentation on this. So was wondering is lazy loading supported? How the search works in general, for the visible records, only or?

kamorel (approved):

frontend/src/components/object/ObjectTable.vue
jujaga: `FilterMatchMode` component should be re-wrapped in our `@/lib/primevue` module.

jujaga (commented):
Looks functional at semantic and visual inspection level. Let's put in some // TODO notes in code so we remember to revisit when we update primevue dependency as well as re-wrap the new component in our lib.

frontend/src/components/object/ObjectTable.vue
loneil: @dc-bcgov yeah the primevue data search or filter is just looking through what is in the table. So API lookup based on input from that field is not supported in this cut of BCBox.
https://primevue.org/datatable/#filter

@jujaga fixed. Different export format for that API file

PR #64, opened 2023-04-03
Home page

kamorel:
# Description

Implements the new home page as designed.

## Types of changes

## Checklist


frontend/src/views/HomeView.vue
```diff
@@ -17,16 +31,184 @@ const { getConfig } = storeToRefs(useConfigStore());
...
+          Log in to get started
+        </Button>
+      </h3>
+      <h3 v-else>
+        <Button
+          @click="goToBuckets()"
+        >
+          Go to My Buckets
+        </Button>
+      </h3>
```
jujaga, line 61: The button action for both of these should be the same: goToBuckets(). We shouldn't be permuting its behavior here because we should be depending on underlying infra to do the login for us if necessary. The only thing that should be changing here is the flavor text, and that can be a simple computed based on whether you're actually authenticated or not.

Another note is that for link navigation instances like this, we should be leveraging `<router-link>` instead. This could be collapsed down to something like the following IIRC:
```vue
<router-link to="/list/buckets">
  <Button>
    {{ !getIsAuthenticated ? 'Log in to get started' : 'Go to My Buckets' }}
  </Button>
</router-link>
```

frontend/src/views/HomeView.vue
```diff
@@ -17,16 +31,184 @@ const { getConfig } = storeToRefs(useConfigStore());
...
+    >
+      <p class="text-xl">
+        We continue to expand the features of BCBox. Stay tuned for documentation on our wiki.<br /><br />
+      </p>
+    </div>
+    <div class="flex justify-content-center mb-8 text-xl">
+      <ul>
+        <li>
+          Need a bucket? Contact <a href="mailto: NRIDS.Optimize@gov.bc.ca">NRIDS Optimization</a>
+          (Natural Resource ministries) or your <br /> ministry's service desk
```
jujaga, line 136: Fine for now, but we should put a ticket in at some point in the future to parameterize this in the /config endpoint.

frontend/src/views/HomeView.vue
```diff
@@ -17,16 +31,184 @@ const { getConfig } = storeToRefs(useConfigStore());
...
+        <li>
+          Learn more about the
+          <a
+            href="https://bcgov.github.io/common-service-showcase/"
+            target="_blank"
+          >Common Services Showcase Team</a>
+        </li>
+        <li>
+          Email the <a href="mailto: nrm.commonserviceshowcase@gov.bc.ca">Common Services Showcase Team</a>
+          with questions or feedback
```
jujaga, line 154: Fine for now, but we should put a ticket in at some point in the future to parameterize this in the /config endpoint.

frontend/src/views/HomeView.vue
```diff
@@ -17,16 +31,184 @@ const { getConfig } = storeToRefs(useConfigStore());
...
+          Log in to get started
+        </Button>
+      </h3>
+      <h3 v-else>
+        <Button
+          @click="goToBuckets()"
+        >
+          Go to My Buckets
+        </Button>
+      </h3>
```
jujaga, line 172: The button action for both of these should be the same: goToBuckets(). We shouldn't be permuting its behavior here because we should be depending on underlying infra to do the login for us if necessary. The only thing that should be changing here is the flavor text, and that can be a simple computed based on whether you're actually authenticated or not.

Another note is that for link navigation instances like this, we should be leveraging `<router-link>` instead. This could be collapsed down to something like the following IIRC:
```vue
<router-link to="/list/buckets">
  <Button>
    {{ !getIsAuthenticated ? 'Log in to get started' : 'Go to My Buckets' }}
  </Button>
</router-link>
```

jujaga (approved):

dc-bcgov (commented):
Important: I think this A/C is missing: "Ensure the Header has a "Beta" tag on it for now". Tested and couldn't see it when I was logged in and logged out.

Less important: I was wondering, at the line 41 'HomeView.vue' if we want to read the text "Go to my buckets" and another one, from config file instead? We might change the text in the future.

kamorel:
> 

The warning banner should cover the fact that the application is under development.

My Buckets is a proper name and will not be changed unless the design calls for the navigation links to change.

PR #66, opened 2023-04-05
Check bucket exists before create/update

kamorel:
# Description

Creates a form differential so COMS only receives the modified bucket information for updates.

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/utils/utils.ts
+++ b/frontend/src/utils/utils.ts
@@ -1,4 +1,23 @@
 /**
+ * @function differential
+ * Create a key/value differential from source against comparer
+ * @param {object} source Source object
+ * @param {object} comparer The object to be compared against
+ * @returns {object} Object containing the non-matching key/value pairs in the source object
+ */
+export function differential(source: any, comparer: any) {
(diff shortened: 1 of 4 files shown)


frontend/src/utils/utils.ts
```diff
@@ -1,3 +1,22 @@
...
+  const diff: any = {};
+  for (const [key, value] of Object.entries(source)) {
+    if (!Object.prototype.hasOwnProperty.call(comparer, key) ||
+      (Object.prototype.hasOwnProperty.call(comparer, key) && comparer[key] !== value)) {
+      diff[key] = value;
+    }
+  }
+
+  return diff;
+}
```
jujaga, line 18: This can be simplified by leveraging `Object.fromEntries`
```ts
export function differential(source: any, comparer: any) {
  return Object.fromEntries(Object.entries(source)
    .filter(([key, value]) => comparer.hasOwnProperty(key) && comparer[key] !== value)
  );
}
```

jujaga (approved):

PR #73, opened 2023-04-24
Store unit tests

kamorel:
# Description

Adds basic tests for
- AppStore
- BucketStore
- ConfigStore
- MetadataStore
- ObjectStore
- TagStore
- UserStore

PrimeVue `useToast` has been wrapped to not be included during unit testing as it was impossible to mock.

Some tests include basic error checking by monitoring the calls to the toast. Some are skipped because they are showing the incorrect number of calls to them. Gotta figure out why.

Slipstreams a fix to remove dev dependencies in the build process.

## Types of changes

## Checklist


frontend/tests/unit/store/appStore.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+describe('App Store', () => {
+  beforeEach(() => {
+    // Creates a fresh pinia and make it active so it's automatically picked
+    // up by any useStore() call without having to pass it to it:
+    // `useStore(pinia)`
+    setActivePinia(createPinia());
+  });
+
+  it('beginDeterminateLoading', () => {
+    const store = useAppStore();
```
dc-bcgov, line 13: I see 'useAppStore()' has been used at each 'it', can we put it in beforeEach?

frontend/tests/unit/store/tagStore.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+});
+
+describe('Config Store', () => {
+
+  describe('fetchTagging', () => {
+    it('fetches the tags', async () => {
+      const appStore = useAppStore();
+      const tagStore = useTagStore();
+
+      const tag: Tagging = {
```
dc-bcgov, line 28: As 'tag' variable has been used 4 times in the test, it might be better for 'tag' to be defined at the top level (line 11 for example) rather than to be function nested.

jujaga (approved):

TimCsaky (approved):
seems like you figured out a lot of stuff here. great work.

PR #79, opened 2023-05-02
Object version details

kamorel:
# Description

Restructures the object details view to allow a versionId parameter. Data will be loaded for the given version. If the parameter is not defined it will default to the latest version.

## Types of changes

## Checklist


frontend/src/components/object/ObjectFileDetails.vue
```diff
@@ -42,6 +55,9 @@ const permissionsObjectId: Ref<string> = ref('');
 const permissionsObjectName: Ref<string> = ref('');
 const obj: Ref<COMSObject | undefined> = ref(undefined);
 const bucketId: Ref<string> = ref('');
+const latestVersionId: Ref<string | undefined> = ref(undefined);
+const version: Ref<Version | undefined> = ref(undefined);
+const filename: Ref<string | undefined> = ref(undefined);
```
jujaga, line 60: Ideally would be good to keep these alphabetized if possible.

frontend/src/components/object/ObjectFileDetails.vue
```diff
@@ -81,68 +97,113 @@ onBeforeMount( async () => {
...
-        <DeleteObjectButton
-          v-if="permissionStore.isObjectActionAllowed(
-            props.objectId, getUserId, Permissions.DELETE, bucketId)"
-          :mode="ButtonMode.ICON"
-          :ids="[props.objectId]"
-          @on-deleted-success="onDeletedSuccess"
+      </div>
+      <div>
+        <Divider layout="vertical" />
+      </div>
```
jujaga, line 193: Does this have to be in its own div wrapper or can it standalone without the wrapper?

frontend/src/components/object/ObjectVersion.vue
```diff
@@ -0,0 +1,174 @@
...
+    summary: 'Success',
+    detail: 'File deleted',
+    life: 3000
+  });
+}
+
+async function load() {
+  await versionStore.fetchVersions({objectId: props.objectId});
+  const versions = versionStore.findVersionsByObjectId(props.objectId);
+  await userStore.fetchUsers({userId:[versions.map( (x: Version) => x.createdBy)]});
```
jujaga, line 61: Style: spacing/formatting seems off here.

frontend/src/components/object/ObjectVersion.vue
```diff
@@ -0,0 +1,174 @@
...
+
+  tableData.value = versions.map( (v: Version) => {
+    return {
+      ...v,
+      createdByName: getUserSearch.value.find( (u: User) => u.userId === v.createdBy )?.fullName
+    };
+  });
+}
```
jujaga, line 69: Using a closure you can avoid needing to write the return:
```ts
  tableData.value = versions.map( (v: Version) => ({
    ...v,
    createdByName: getUserSearch.value.find( (u: User) => u.userId === v.createdBy )?.fullName
  }));
```
Generally recommend favoring this form of syntax over an explicit return unless there needs to be operations that have to be done prior to returning out of the mapping.

frontend/src/components/object/ObjectVersion.vue
```diff
@@ -0,0 +1,174 @@
...
+              <font-awesome-icon icon="fa-solid fa-circle-info" />
+            </Button>
+            <DeleteObjectButton
+              v-if="permissionStore.isObjectActionAllowed(
+                data.id, getUserId, Permissions.DELETE, props.bucketId as string)"
+              :mode="ButtonMode.ICON"
+              :ids="['TODO']"
+              @on-deleted-success="onDeletedSuccess"
+            />
+          </template>
```
jujaga, line 169: As these button actions are not yet wired in, we should temporarily hardcode these as disabled so that if this does show up in prod, people aren't going to trip over incomplete functionality.

frontend/src/utils/constants.ts
```diff
@@ -2,7 +2,7 @@ export const BucketConfig = Object.freeze({
   HEADER_NEW_BUCKET: 'Configure bucket',
   TITLE_NEW_BUCKET: 'Use this form to configure a bucket to be used in BCBox for the first time.',
   NOTICE_NEW_BUCKET: 'If you intend to share files in your bucket with BCeID users, you are required'
-  + ' to email IDIM.Consulting@gov.bc.ca to share your intentions and where you intend to advertise this.'
+    + ' to email IDIM.Consulting@gov.bc.ca to share your intentions and where you intend to advertise this.'
```
jujaga, line 5: At some point we're going to need to revisit this and have this be driven from the /config endpoint. We should put in a TODO comment here so that we don't forget to address this down the line.

frontend/src/components/object/ObjectFileDetails.vue
```diff
@@ -81,68 +97,111 @@ onBeforeMount( async () => {
   }
 });
 
-watch( [props, getObjects], () => {
-  metadataStore.fetchMetadata({objectId: props.objectId });
-  tagStore.fetchTagging({objectId: props.objectId });
+watch( [props, getObjects], async () => {
+  await metadataStore.fetchMetadata({objectId: props.objectId});
```
TimCsaky, line 101: Just wondering if having two metadata stores is the best pattern.
maybe a single metadata store could have actions that inherit object > version data from the other stores.
I think you already have the object > version relation in the metadataStore in versioStore.ts (findLatestVersionIdByObjectId)
.. so you can get metadata for an object by looking at the latest version, instead of calling a separate action everywhere.
But maybe that pattern has its drawbacks too. I think what you have is fine.

TimCsaky (commented):
Awesome work.. youve done a lot of stuff here.. and the store is proving to be the core issue again i guess.

frontend/src/components/object/ObjectTable.vue
```diff
@@ -204,7 +204,7 @@ const filters = ref({
         :hidden="props.objectInfoId ? true : false"
       >
         <template #body="{ data }">
-          {{ formatDateLong(data.updatedAt) }}
+          {{ data.updatedAt ? formatDateLong(data.updatedAt) : formatDateLong(data.createdAt) }}
```
jujaga, line 207: Can change this to
```ts
formatDateLong(data.updatedAt ?? data.createdAt)
```
To avoid needing to use an explicit branching ternary.

frontend/src/components/object/ObjectVersion.vue
```diff
@@ -117,10 +117,10 @@ watch( [props], () => {
                 :to="{ name: RouteNames.DETAIL_OBJECTS,
                        query: { objectId: props.objectId, versionId: data.id } }"
               >
-                {{ formatDateLong(data.updatedAt) }}
+                {{ data.updatedAt ? formatDateLong(data.updatedAt) : formatDateLong(data.createdAt) }}
               </router-link>
               <span v-else>
-                {{ formatDateLong(data.updatedAt) }}
+                {{ data.updatedAt ? formatDateLong(data.updatedAt) : formatDateLong(data.createdAt) }}
```
jujaga:
```ts
formatDateLong(data.updatedAt ?? data.createdAt)
```

jujaga (approved):

frontend/src/components/object/ObjectVersion.vue
```diff
@@ -0,0 +1,175 @@
...
+// State
+const tableData: Ref<Array<VersionDataSource>> = ref([]);
+
+// Actions
+const toast = useToast();
+
+// eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
+const showInfo = async (id: string) => {
+  // emit('show-object-info', id);
+};
```
jujaga, line 47: Do we still need this function at all if it's effectively a no-op?

jujaga (approved):
Looks good from what I could tell for the most part. Just some very minor style notes but good otherwise. ✅

frontend/src/components/object/ObjectFileDetails.vue
bcdcgv: Agree with Tim.

frontend/src/components/object/ObjectVersion.vue
kamorel: Leaving in for now. The button that even calls this is disabled, but I believe I will be making use of this in the next PR for object updating.

bcdcgv (approved):

PR #82, opened 2023-05-11
Implement loading icon or hover effect

bcgv123:
# Description

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3167

Hover effect: css has been used as the last tool, as font-awesome-icons do not support hover animation.
Loading icon: primevue spinner has been used for Bucket and Object tables. 

**Fix after Tyler's comment:**
Screenshot 2023-05-12 at 3 06 08 PM

## Checklist


frontend/src/components/bucket/BucketTable.vue
```diff
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import ProgressSpinner from 'primevue/progressspinner';
```
TimCsaky, line 2: i would create our own 'Loader' component in `/src/components/layout`. 
import it into AppLayout.vue .. so it wont need importing again in child components.
eg in datatables you can do `<template #loading><Loader /></template>`
and in `OidcLoginView.vue` you can replace `<ProgressSpinner />` with `<Loader />`. 
what do you think guys?

TimCsaky: also in Loader we can customise the look with something like this:
```css
.bcbox-loading { 
 background: white;
 opacity: .3;

 .p-progress-spinner { 
  position: absolute;
  top: 30%;
  left: 50%;
  transform: translate(-50%, -50%);
  opacity: 1; 
 }
}
```
see: https://primevue.org/progressspinner/#style

jujaga: I would say it's worth encapsulating into its own SFC as you recommend if and only if it has something we're embedding in as a part of the SFC definition like some default scss styling.
Also note that we should be re-routing the progress spinner through the @lib wrapper that we have instead of directly importing.

frontend/src/assets/main.scss
```diff
@@ -124,3 +124,9 @@ td .wrap-block{ width: inherit; }
     text-transform: none;
   }
 }
+
+.p-button {
+  &:hover {
+    opacity: 0.8;
```
jujaga, line 130: Is an opacity change what the public slider/toggle button is doing under the hood? If it is, let's do it this way; otherwise I'd recommend doing whatever that component is doing under the hood in lieu of this.

bcdcgv: Screenshot 2023-05-12 at 8 19 08 AM

If not wrong, it seems like toggle hover is implemented by hardcoding background colour. Do we want to go this way?

tpantella (commented):
Reviewed using https://bcbox-dev-pr-82.apps.silver.devops.gov.bc.ca/ as provided by Dragan.

Hover appears good, a spot test around various actions seem appropriately subtle as a slight difference in lightness/darkness, agree with the dev choice to lighten darker elements on hover.

Loading animation of table appears to be red circle. Red usually indicates failure and is not a prominent colour in our scheme, not sure if this is an error or intentional but loading should be a shade of blue or grey for consistency. Table also briefly loads with large dark grey dectangle, would prefer a subtle loading colour more congruent with the design such as a white.

Attaching image I was able to capture of red spinner, it's tough to get because it's so quick so you only see part of it.

PR Hover Loading Review TP

bcdcgv:
> Reviewed using https://bcbox-dev-pr-82.apps.silver.devops.gov.bc.ca/ as provided by Dragan.
> 
> Hover appears good, a spot test around various actions seem appropriately subtle as a slight difference in lightness/darkness, agree with the dev choice to lighten darker elements on hover.
> 
> Loading animation of table appears to be red circle. Red usually indicates failure and is not a prominent colour in our scheme, not sure if this is an error or intentional but loading should be a shade of blue or grey for consistency. Table also briefly loads with large dark grey dectangle, would prefer a subtle loading colour more congruent with the design such as a white.
> 
> Attaching image I was able to capture of red spinner, it's tough to get because it's so quick so you only see part of it.
> 
> PR Hover Loading Review TP

As requested, overwriting spinner colour to grey and loading overlay to white.

frontend/src/assets/main.scss
```diff
@@ -124,3 +124,22 @@ td .wrap-block{ width: inherit; }
...
```
jujaga, line 130:
```css
text-decoration: underline;
```
This needs to be added in to be compliant with BCGov Style Guide.

frontend/src/components/layout/Loader.vue
```diff
@@ -0,0 +1,14 @@
+<script setup lang="ts">
+import { ProgressSpinner } from '@/lib/primevue';
+
+</script>
+
+<template>
+  <ProgressSpinner />
```
TimCsaky, line 7: to change the style you can do this kind of thing:
```html
<ProgressSpinner style="width: 50px; height: 50px" strokeWidth="4" />
 ```

File: frontend/src/assets/main.scss [hunk 5]
```diff
@@ -124,3 +124,22 @@ td .wrap-block{ width: inherit; }
     text-transform: none;
   }
 }
+
+.p-button {
```
TimCsaky, line 128: to group the selectors, you can change this line to:
`.p-button, .p-inputswitch-slider {`
and you can remove the new style you added in line 134

kamorel (approved):

jujaga (approved):

PR #83, opened 2023-05-15
New version uploading

kamorel:
# Description

Allows new version of objects to be uploaded via a single file selection upload button in the object details view.

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/object/ObjectUploadBasic.vue
+++ b/frontend/src/components/object/ObjectUploadBasic.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { ref } from 'vue';
 
-import { Button, useToast } from '@/lib/primevue';
+import { Button, useConfirm, useToast } from '@/lib/primevue';
 import { useObjectStore } from '@/store';
 
 import type { Ref } from 'vue';
@@ -14,6 +14,9 @@ type Props = {
 
 const props = withDefaults(defineProps<Props>(), {});
 
+// Emits
+const emit = defineEmits(['on-file-changed', 'on-file-uploaded']);
+
 // Store
 const objectStore = useObjectStore();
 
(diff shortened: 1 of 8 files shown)


frontend/src/components/object/ObjectUploadBasic.vue
```diff
@@ -0,0 +1,88 @@
...
+
+const confirmUpdate = () => {
+  confirm.require({
+    message: 'Please confirm that you want to upload a new version. All other versions will still be stored.',
+    header: 'Upload new version',
+    acceptLabel: 'Confirm',
+    rejectLabel: 'Cancel',
+    accept: async () => {
+      onUpload();
+    },
```
jujaga, line 39: this could just be
```ts
accept: onUpload();
```

frontend/src/components/object/ObjectUploadBasic.vue
```diff
@@ -0,0 +1,88 @@
...
+    outlined
+    @click="onSelectFile"
+  >
+    <font-awesome-icon
+      icon="fa-solid fa-upload"
+      class="mr-1"
+    />
+    Upload new version
+  </Button>
+  <input
```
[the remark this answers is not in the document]
jujaga: I second this. This gets the job done for now, but we really need to consider how to unify and asynchronize ongoing upload progress from page state at some point soon.

frontend/src/services/objectService.ts
```diff
@@ -124,12 +124,15 @@ export default {
   /**
    * @function updateObject
    * Update the object record (will add new version)
+   * @param {String} objectId The id for the object
+   * @param {Any} object Object to be created
```
jujaga, line 128: IIRC these types should be lowercase - we don't use the boxed string type, and any in caps doesn't exist.

frontend/src/store/objectStore.ts
```diff
@@ -166,6 +166,19 @@ export const useObjectStore = defineStore('object', () => {
     }
   }
 
+  async function updateObject(object: any, objectId: string, axiosOptions?: AxiosRequestConfig) {
```
jujaga, line 169: Don't forget to eventually put in JSDocs for these functions over time.

jujaga (commented):
Would be nice to see if we can have the button be sharing the same div as the header - the offset is a bit jarring as shown below:
Screenshot 2023-05-15 at 15-11-27 BCBox
We'll also need to pretty soon revisit the network calls being done. We shouldn't be dispatching a huge array of objectId requests on this page... ever, but this would need a bit more investigation with network call management in the store.

jujaga:
Subsumed by #84

PR #84, opened 2023-05-16
Version downloading/deleting

kamorel:
# Description

Allows specific object versions to be downloaded and deleted.

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/object/DeleteObjectButton.vue
+++ b/frontend/src/components/object/DeleteObjectButton.vue
@@ -18,7 +18,6 @@ const props = withDefaults(defineProps<Props>(), {});
 // Emits
 const emit = defineEmits(['on-deleted-success', 'on-deleted-error']);
 
-
 // Store
 const objectStore = useObjectStore();
 
--- a/frontend/src/components/object/DeleteObjectButton.vue
+++ b/frontend/src/components/object/DeleteObjectButton.vue
@@ -43,7 +43,7 @@ const confirmDelete = () => {
       accept: async () => {
         try {
           await objectStore.deleteObjects(props.ids, props.versionId);
-          emit('on-deleted-success');
+          emit('on-deleted-success', props.versionId);
         } catch (error: any) {
           toast.error(`Error deleting one or more ${item}s`);
           emit('on-deleted-error');
--- a/frontend/src/components/object/DeleteObjectButton.vue
+++ b/frontend/src/components/object/DeleteObjectButton.vue
@@ -11,9 +11,12 @@ type Props = {
   disabled?: boolean;
   ids: Array<string>;
   mode: ButtonMode;
+  versionId?: string; // Only use this when deleting a single object
 };
 
-const props = withDefaults(defineProps<Props>(), {});
+const props = withDefaults(defineProps<Props>(), {
+  versionId: undefined
+});
 
 // Emits
 const emit = defineEmits(['on-deleted-success', 'on-deleted-error']);
@@ -32,13 +35,13 @@ const confirmDelete = () => {
   if (props.ids.length) {
     const msgContext = props.ids.length > 1 ? `the selected ${props.ids.length} files` : 'this file';
     confirm.require({
-      message: `Please confirm that you want to delete ${msgContext}`,
+      message: `Please confirm that you want to delete ${msgContext}.`,
       header: `Delete ${props.ids.length > 1 ? 'items' : 'item'}`,
       acceptLabel: 'Confirm',
       rejectLabel: 'Cancel',
       accept: async () => {
         try {
-          await objectStore.deleteObjects(props.ids);
+          await objectStore.deleteObjects(props.ids, props.versionId);
           emit('on-deleted-success');
         } catch (error: any) {
           toast.error('Error deleting one or more files');
--- a/frontend/src/components/object/DeleteObjectButton.vue
+++ b/frontend/src/components/object/DeleteObjectButton.vue
@@ -33,10 +33,11 @@ const toast = useToast();
 
 const confirmDelete = () => {
   if (props.ids.length) {
-    const msgContext = props.ids.length > 1 ? `the selected ${props.ids.length} files` : 'this file';
+    const item = props.versionId ? 'version' : 'object';
+    const msgContext = props.ids.length > 1 ? `the selected ${props.ids.length} ${item}s` : `this ${item}`;
     confirm.require({
       message: `Please confirm that you want to delete ${msgContext}.`,
-      header: `Delete ${props.ids.length > 1 ? 'items' : 'item'}`,
+      header: `Delete ${props.ids.length > 1 ? item + 's' : item }`,
       acceptLabel: 'Confirm',
       rejectLabel: 'Cancel',
       accept: async () => {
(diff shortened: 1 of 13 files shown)


frontend/src/components/object/ObjectFileDetails.vue
```diff
@@ -74,6 +75,18 @@ function onDeletedSuccess() {
...
+  await objectStore.fetchObjects({objectId: props.objectId, userId: getUserId.value, bucketPerms: true});
+  await versionStore.fetchVersions({ objectId: props.objectId });
+
+  // Obtaining the version id and passing it to the route forces a destruct/construct of the component
+  // Easier approach than attempting to in-place refresh all the data
+  router.push({ name: RouteNames.DETAIL_OBJECTS, query: {
+    objectId: props.objectId,
+    versionId: versionStore.findLatestVersionIdByObjectId(props.objectId)
+  }});
+}
```
jujaga, line 88: I can't quite remember, but there may be a way to only update a certain set of sub-components by force updating its id key and then having that component re-render. What you have here should work, and also properly addresses the url location though.

frontend/src/components/object/ObjectUploadBasic.vue
```diff
@@ -0,0 +1,86 @@
...
+const confirmUpdate = () => {
+  confirm.require({
+    message: 'Please confirm that you want to upload a new version.',
+    header: 'Upload new version',
+    acceptLabel: 'Confirm',
+    rejectLabel: 'Cancel',
+    accept: () => onUpload(),
+    reject: () => {
+      // Intentionally left empty
+    },
```
jujaga, line 40: Does this attribute need to be defined with an empty callback? Would be ideal to just not have it even defined at all if possible.

frontend/src/components/object/DeleteObjectButton.vue
```diff
@@ -33,13 +35,13 @@ const confirmDelete = () => {
   if (props.ids.length) {
     const msgContext = props.ids.length > 1 ? `the selected ${props.ids.length} files` : 'this file';
```
TimCsaky, line 36: feels like the confirm message should specify if its a version or an object being deleted. you could do:
```js
const item = props.versionId ? 'version' : 'object';  
const msgContext = props.ids.length > 1 ? `the selected ${props.ids.length} ${item}s` : 'this ${item}';
header: `Delete ${props.ids.length > 1 ? item + 's' : item }`,
...
toast.error('Error deleting one or more ' + item + 's');
```
i suspect when we look at doing hard/soft deletes this will need to be re-visited

frontend/src/components/object/ObjectUploadBasic.vue
```diff
@@ -0,0 +1,86 @@
...
+  });
+};
+
+const onChange = async (event: any) => {
```
jujaga, line 44: Not async.

frontend/src/components/object/ObjectUploadBasic.vue
```diff
@@ -0,0 +1,86 @@
...
+  file.value = event.target.files[0];
+  emit('on-file-changed');
+  confirmUpdate();
+};
+
+const onSelectFile = async () => {
```
jujaga, line 50: Not async.

frontend/src/components/object/ObjectVersion.vue
```diff
@@ -39,15 +40,18 @@ const { getUserSearch } = storeToRefs(userStore);
...
 
-// eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
-const showInfo = async (id: string) => {
-  // emit('show-object-info', id);
-};
-
-function onDeletedSuccess() {
+async function onDeletedSuccess() {
   toast.success('File deleted');
+  await versionStore.fetchVersions({ objectId: props.objectId });
```
TimCsaky, line 48: i think i found when deleting any version except latest, that version remains in the table until you do a hard refresh.
maybe the fetchVersions() needs to do that partition logic somewhere to remove that old version from the store (?)

TimCsaky: or maybe it's because the route isnt changing 🤔

frontend/src/components/object/ObjectUploadBasic.vue
```diff
@@ -0,0 +1,86 @@
...
+    />
+    Upload new version
+  </Button>
+  <input
+    ref="fileInput"
+    type="file"
+    style="display: none"
+    accept="*"
+    @change="onChange"
+  />
```
TimCsaky, line 85: FYI. Sharolyn mentioned that a 'uploading progress indicator' would be good. 
i would have thought it would be ok to tackle that later, along with showing a something similar in the main upload panel. realistically people aren't going to be updating big files anytime soon.
... but, if we did want to do something short term.. maybe we could put in something here like:
`<Spinner v-if"!done" />` (using a flag in the onUpload() function)

frontend/src/components/object/ObjectVersion.vue
```diff
@@ -142,24 +146,28 @@ watch( [props], () => {
               v-if="data.public || permissionStore.isObjectActionAllowed(
```
TimCsaky, line 146: just a thought..i think they can't even get to the file details page if they havent got READ permission. but if we re-use this component, then this is good.

frontend/src/components/object/ObjectFileDetails.vue
```diff
@@ -185,6 +199,13 @@ watch( [props, getObjects], async () => {
       </div>
       <Divider layout="vertical" />
       <div class="flex flex-column w-6 gap-3 py-5">
+        <div class="flex flex-row-reverse">
+          <ObjectUploadBasic
```
TimCsaky, line 203: wondering if we should be checking for UPDATE permission before showing this button

frontend/src/components/object/ObjectMetadata.vue
```diff
@@ -53,36 +53,11 @@ watch([props, tsGetMetadata,vsGetMetadata] , () => {
...
-        >
-          <template #body="{ data }">
-            <div class="wrap-block">
-              {{ data.value }}
-            </div>
-          </template>
-        </Column>
-      </DataTable>
-    </div>
+    <GridRow
```
TimCsaky, line 56: i think this is good.. not sure datatable was neccessary

frontend/src/components/object/ObjectVersion.vue
```diff
@@ -39,15 +40,18 @@ const { getUserSearch } = storeToRefs(userStore);
...
-// eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
-const showInfo = async (id: string) => {
-  // emit('show-object-info', id);
-};
-
-function onDeletedSuccess() {
+async function onDeletedSuccess() {
   toast.success('File deleted');
+  await versionStore.fetchVersions({ objectId: props.objectId });
+  router.push({ name: RouteNames.DETAIL_OBJECTS,
```
TimCsaky, line 49: this is good for when youre deleting the version you are on (viewing with versionId in url) 
but if youre deleting any other version , it doesnt seem to remove it from table (without a hard refresh) and doesn't really need to navigate to the latest.

maybe this would work:
```js
// if currently on latest version, re-load route
if (props.versionId === versionStore.findLatestVersionIdByObjectId(props.objectId) {
  router.push(...
 ```js

File: frontend/src/components/object/ObjectVersion.vue [hunk 12]
```diff
@@ -39,15 +40,18 @@ const { getUserSearch } = storeToRefs(userStore);
...
+const router = useRouter();
 const toast = useToast();
 
-// eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
-const showInfo = async (id: string) => {
-  // emit('show-object-info', id);
-};
-
-function onDeletedSuccess() {
+async function onDeletedSuccess() {
```
TimCsaky, line 46: maybe passing objectId and versionId as params of this function would help.
not sure the pattern for emitting that

jujaga (approved):

TimCsaky (approved):

PR #85, opened 2023-05-23
Navigation notifications during uploads

kamorel:
# Description

During object uploads
1. shows a custom message confirmation during navigation away from upload components
2. shows a browser default confirmation during window navigation (refresh, tab close, etc)

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/object/ObjectUpload.vue
+++ b/frontend/src/components/object/ObjectUpload.vue
@@ -49,8 +49,9 @@ const onUpload = async (event: any) => {
     await Promise.allSettled(
       event.files.map(async (file: File) => {
         try {
-          // Infinite timeout for big files upload to avoid timeout error
           appStore.beginUploading();
+
+          // Infinite timeout for big files upload to avoid timeout error
           await objectStore.createObject(file, bucketId, { timeout: 0 });
           successfulFiles.value.push(file);
         } catch (error: any) {
--- a/frontend/src/components/object/ObjectUpload.vue
+++ b/frontend/src/components/object/ObjectUpload.vue
@@ -4,7 +4,7 @@ import { ref } from 'vue';
 
 import ObjectUploadFile from '@/components/object/ObjectUploadFile.vue';
 import { Button, FileUpload, useToast } from '@/lib/primevue';
-import { useAuthStore, useObjectStore } from '@/store';
+import { useAuthStore, useAppStore, useObjectStore } from '@/store';
 
 import type { Ref } from 'vue';
 
@@ -19,6 +19,7 @@ const props = withDefaults(defineProps<Props>(), {
 });
 
 // Store
+const appStore = useAppStore();
 const objectStore = useObjectStore();
 const { getUserId } = storeToRefs(useAuthStore());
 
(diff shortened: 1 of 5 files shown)


frontend/src/components/object/ObjectUploadBasic.vue
```diff
@@ -51,12 +52,20 @@ const onSelectFile = () => {
 const onUpload = async () => {
   try {
     if( file.value ) {
+      toast.info('Upload starting...');
```
TimCsaky, line 55: i noticed there's a 'sticky' option for the toast popups. might be able to close it once the upload has finished.

frontend/src/components/object/ObjectUpload.vue
```diff
@@ -42,17 +43,24 @@ const onUpload = async (event: any) => {
     successfulFiles.value = [];
     failedFiles.value = [];
 
+    toast.info('Upload starting...');
```
jujaga, line 46: File upload starting...

frontend/src/components/object/ObjectUploadBasic.vue
jujaga: File upload starting...

I think a sticky toast option only makes sense if and only if the toast message contents were also dynamic in nature (aka seeing progress, indeterminate, etc). Otherwise having it just stay there is not likely to be a great user experience, especially if multiple files are being uploaded at once.

frontend/src/router/index.ts
```diff
@@ -140,15 +140,22 @@ export default function getRouter() {
...
 
+
+    // Uploading navigation guard
+    if (appStore.getIsUploading) {
+      if (!confirm('Navigation may cancel upload(s) in progress. ' +
+        'Please confirm you want to navigate from current page.')) {
+        return false;
+      }
+    }
+
```
jujaga, line 151: I'm a bit unsure of order of operations and lifecycle here. Should this be in the `beforeRouteLeave` stage of the lifecycle instead?

frontend/src/router/index.ts
```diff
@@ -140,15 +140,22 @@ export default function getRouter() {
...
     // Authentication Guard
     if (to.meta.requiresAuth) {
       const user = await authService.getUser();
       if (!user || user.expired) {
         router.replace({ name: RouteNames.LOGIN });
       }
     }
-
-    next();
```
jujaga, line 151: We shouldn't be changing the signature of the beforeEach lifecycle to be ignoring the next(). This callback should always be there to ensure that any subsequent lifecycle events are being invoked. By removing this, later lifecycle declarations many not ever be executed.

frontend/src/store/appStore.ts
```diff
@@ -25,6 +27,7 @@ export const useAppStore = defineStore('app', () => {
     getLoadingCalls: computed(() => state.loadingCalls.value),
     getLoadingMode: computed(() => state.loadingMode.value),
     getLoadingValue: computed(() => state.loadingValue.value),
+    getIsUploading: computed(() => state.uploadingCalls.value > 0),
```
jujaga, line 30: Alphabetize

jujaga (approved):

TimCsaky (approved):

PR #89, opened 2023-06-08
Initial Vue SFC unit test patterns

kamorel:
# Description

Adds a few small SFC tests to show how to correctly mount components with proper plugins and initial store states, access various parts of the DOM, and router mocking and testing.

## Types of changes

## Checklist

### DIFF
--- a/frontend/tests/unit/components/form/TextInput.spec.ts
+++ b/frontend/tests/unit/components/form/TextInput.spec.ts
@@ -4,7 +4,7 @@ import PrimeVue from 'primevue/config';
 import TextInput from '@/components/form/TextInput.vue';
 
 describe('TextInput.vue', () => {
-  it('renders', async () => {
+  it('renders', () => {
     const wrapper = shallowMount(TextInput, {
       props: {
         name: 'test',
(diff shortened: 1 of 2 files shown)


frontend/tests/unit/components/form/TextInput.spec.ts
```diff
@@ -1,32 +1,79 @@
...
+        name: 'test',
+        placeholder: 'some placeholder text'
+      },
+      global: {
+        plugins: [PrimeVue]
+      }
+    });
+
+    const input = wrapper.getComponent({ name: 'InputText' });
+    expect(input.attributes('placeholder')).toBe('some placeholder text');
```
jujaga, line 62: As a mental note, it may be worth letting the toBe value be assigned as a constant like placeholderValue, before the const wrapper declaration. That way, when we do the shallowMount and do a validation, we can reference back on the same constant and know that it's the same value and memory location, instead of copy-pasting the same string in two places.

jujaga (approved):
From what I can gather I think this looks good. Please do one pass through to make sure that the it callback body functions do not have async on them unless they're absolutely needed; most of them do not have an explicit await statement in them, leading me to believe that most of these can be declared synchronously instead. Tentatively approving. 👍

TimCsaky (approved):

PR #90, opened 2023-06-14
Metadata & tags on upload

kamorel:
# Description

Allows the user to attach metadata and tags (up to 9, to allow coms-id), via a modal form prior to object upload.

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/object/ObjectUploadFile.vue
+++ b/frontend/src/components/object/ObjectUploadFile.vue
@@ -10,6 +10,7 @@ import type { ObjectMetadataTagFormType } from '@/components/object/ObjectMetada
 
 // Props
 type Props = {
+  editable?: boolean;
   files: Array<any>; // TODO: Change any to more specific type
   badgeProps: any;
   formData?: Array<ObjectMetadataTagFormType>;
@@ -17,6 +18,7 @@ type Props = {
 };
 
 const props = withDefaults(defineProps<Props>(), {
+  editable: false,
   formData: () => []
 });
 
@@ -90,6 +92,7 @@ const closeMetaModal = () => {
       </div>
       <div class="ml-auto">
         <Button
+          v-if="editable"
           class="p-button-lg p-button-rounded p-button-text"
           @click="showMetaModal(file.name)"
         >
--- a/frontend/src/components/object/ObjectUploadFile.vue
+++ b/frontend/src/components/object/ObjectUploadFile.vue
@@ -34,11 +34,9 @@ const formData: Ref<Array<ObjectMetadataTagFormType>> = ref(props.formData);
 
 // Actions
 const showMetaModal = async (filename: string) => {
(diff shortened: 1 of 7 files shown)


frontend/src/components/object/ObjectUploadFile.vue
```diff
@@ -1,16 +1,69 @@
...
+const formData: Ref<Array<ObjectMetadataTagFormType>> = ref(props.formData);
+
+// Actions
+const showMetaModal = async (filename: string) => {
+  const idx = formData.value.findIndex( (x: ObjectMetadataTagFormType) => x.filename === filename);
+
+  metadataTagFormData.value.filename = filename;
+  metadataTagFormData.value.metadata = idx >= 0 ? formData.value[idx].metadata : undefined;
+  metadataTagFormData.value.tagset = idx >= 0 ? formData.value[idx].tagset : undefined;
+
```
TimCsaky, line 42: might be able to simplify lines 37-41 with:
```js
metadataTagFormData.value.metadata = formData.value.find(x => x.filename === filename).metadata;
metadataTagFormData.value.tagset = formData.value.find(x => x.filename === filename).tagset ;
```

TimCsaky (approved):
Looking great.
not sure how.. but we should probably remove or disable the edit icon on the file once it has moved to 'complete'.

frontend/src/services/objectService.ts
```diff
@@ -14,10 +14,40 @@ export default {
...
-      params: { bucketId: bucketId },
+      headers: {
+        'Content-Type': 'multipart/form-data',
+        ...metadata
+      },
+      params: {
+        bucketId: params.bucketId,
+        tagset: tagset
+      }
     };
```
jujaga, line 51: With a bit of clever reordering, you can avoid reassignment hell by reutilizing the config object and reassigning attributes in those values instead. Something like the following would be slightly more efficient and concise:
```ts
const config = {
  headers: { 'Content-Type': 'multipart/form-data' },
  params: { bucketId: bucketId },
};

// Map the metadata if required
if (headers.metadata) {
  config.headers = {
    ...config.headers,
    ...Object.fromEntries((headers.metadata.map((x: { key: string; value: string }) => ([x.key, x.value ]))))
  };
}

// Map the tagset if required
if (params.tagset) {
  config.params.tagset = Object.fromEntries((params.tagset.map((x: { key: string; value: string }) => ([x.key, x.value]))));
}
```

frontend/src/utils/constants.ts
```diff
@@ -45,3 +45,5 @@ export const ToastTimeout = Object.freeze({
   SUCCESS: 3000,
   WARNING: 5000
 });
+
+export const MAX_TAGS = 9;
```
jujaga, line 49: Alphabetical ordering.
Also good to put in a quick JSDoc style comment on why it's 9 (max 10 tags in S3, one always used by coms, etc)

jujaga (approved):

PR #91, opened 2023-06-16
Metadata and tag updating

kamorel:
# Description

Adds ability to update metadata and tags via the object details view.

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/object/ObjectMetadata.vue
+++ b/frontend/src/components/object/ObjectMetadata.vue
@@ -3,20 +3,28 @@ import { storeToRefs } from 'pinia';
 import { onMounted, ref, watch } from 'vue';
 
 import GridRow from '@/components/form/GridRow.vue';
-import { useMetadataStore, useVersionStore } from '@/store';
+import ObjectMetadataTagForm from '@/components/object/ObjectMetadataTagForm.vue';
+import { Button, Dialog, useConfirm } from '@/lib/primevue';
+import { useMetadataStore, useObjectStore, useVersionStore } from '@/store';
 
 import type { Ref } from 'vue';
+import type { ObjectMetadataTagFormType } from '@/components/object/ObjectMetadataTagForm.vue';
 import type { Metadata } from '@/types';
 
 // Props
 type Props = {
+  editable?: boolean;
   objectId: string;
   versionId?: string;
 };
 
 const props = withDefaults(defineProps<Props>(), {
+  editable: true,
   versionId: undefined
 });
+
+// Emits
+const emit = defineEmits(['on-file-uploaded']);
 
 // Store
 const metadataStore = useMetadataStore();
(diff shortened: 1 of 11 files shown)


frontend/src/components/object/ObjectTag.vue
```diff
@@ -2,32 +2,63 @@
...
+const emit = defineEmits(['on-file-uploaded']);
+
 // Store
 const tagStore = useTagStore();
 const versionStore = useVersionStore();
 const { getTagging: tsGetTagging } = storeToRefs(tagStore);
 const { getTagging: vsGetTagging } = storeToRefs(versionStore);
 
 // State
+const metaVisible: Ref<boolean> = ref(false);
```
TimCsaky, line 35: you could follow a convention that works for both metadata and tag
~~showMetaModal~~ showModal
~~closeMetaModal~~ closeModal
~~submitMetaModal~~ submit
~~metaVisible~~ editing
~~metadataTagFormData~~ formData

TimCsaky (approved):
minor comments about the variable naming.. something to consider but not essential.

frontend/src/components/object/ObjectMetadata.vue
```diff
@@ -3,31 +3,75 @@ import { storeToRefs } from 'pinia';
 import { onMounted, ref, watch } from 'vue';
 
 import GridRow from '@/components/form/GridRow.vue';
-import { useMetadataStore, useVersionStore } from '@/store';
+import ObjectMetadataTagForm from '@/components/object/ObjectMetadataTagForm.vue';
```
jujaga, line 6: We have an index.ts directory repackager in /components/object - we should route components through there instead of directly importing the component.

frontend/src/components/object/ObjectMetadataTagForm.vue
```diff
@@ -67,130 +72,134 @@ const onCancel = () => {
...
-              />
+          <div
+            v-for="(meta, index) of fields"
+            :key="'metadata.'+index"
+            class="grid"
+          >
+            <div class="grid col-11">
+              <div class="col">
+                <TextInput
+                  :name="'metadata.'+index+'.key'"
```
jujaga, line 104: We should have spaces between the + for readability purposes. It's easy to miss this concatenation otherwise. Ideally we should try to favor using template strings instead of string concatenation, but unsure if that is the primary intent of assembly here.

frontend/src/components/object/ObjectTag.vue
jujaga: I agree with this - if there is capacity to abstract and decontextualize the component so that it can be reused in multiple contexts, we should position that way.

jujaga (commented):
Will need to ensure the situation where editing from a set with elements to an empty set also works as per earlier conversations.

frontend/src/components/object/ObjectMetadataTagForm.vue
```diff
@@ -67,130 +72,134 @@ const onCancel = () => {
...
-              class="p-button-lg p-button-text p-button-danger p-0"
-              @click="() => { remove(index); if(!fields.length) push({});}"
+              class="p-button p-button-text p-0"
+              @click="push({})"
             >
-              <font-awesome-icon icon="fa-solid fa-minus" />
+              <font-awesome-icon
+                icon="fa-solid fa-plus"
+                class="mr-1"
+              /> Add row
```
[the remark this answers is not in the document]
kamorel: Readability. Does not affect UI

jujaga (approved):

TimCsaky (approved):

loneil (approved):

PR #103, opened 2023-07-12
Tag/meta filter

loneil:
# Description
Add metadata and tag filter. 
Dropdowns call the search endpoints to populate available tags and meta. Then it filters those results by the objectIDs that are actually in the table, as the COMS endpoints will return deleted files and files from other buckets that you have access to.
Possible COMS enhancements to simplify this could be done.

There's a couple UX design things from Tyler's design here we should log as future features or try some more that I was unable to figure out how to coerce into the Primevue multiselect+filter. (but we can pull in this existing bahaviour and enhance it as priorities come).
To consider enhancing (I'll get Koreen to log tickets):
- Don't show anything listed in the dropdowns until something is searched on in the filter
- Only show top 5 results of filtering
- Allow filter to split on delimiter
ie, if you filter `a:b`
it would return all of
`abc:xxbxx`
`a:b`
`123aaaa:567bbb`
rather than just `a:b`

## Types of changes

## Checklist

## Further comments

As we revealed, you can't select multiple of a tag or metadata with the same key. Discussed and going with the option of unchecking any previous filter item that has the same key as the one you are selecting. Might need some additional UX decisions on that down the line. Ticket created to track.


frontend/src/components/object/ObjectFilters.vue
```diff
@@ -0,0 +1,138 @@
...
+  searching.value = true;
+  //await 4 seconds
+  await metadataStore.searchMetadata();
+  searching.value = false;
+};
+
+const searchTagging = async () => {
+  searching.value = true;
+  //await 4 seconds
+  await tagStore.searchTagging();
```
TimCsaky, line 94: I noticed that the dropdown includes tags for all objects, from all buckets in the system (on objects I have READ perm for).

UX question but I suspect we need to filter results of searchTagging to only show tags that are on objects in the current bucket.

You can pass an array of objectId's to the searchTagging endpoint but i dont think that's feasible (too many).
i suspect we need a new bucketId parameter on the COMS `/object/tagging `endpoint. 
but for now maybe we can filter the result of searchTagging against the list of objects in the objectStore (which are already filtered to current bucket, or have a bucketId property). I think applying the tag/metadata filters will filters the list of objects on the objectStore.. so the available tags in the dropdown will also be filtered.. which might be a good user experience.

god this is complicated.. i may have misunderstood all this.. and probably arent making much sense.

TimCsaky (commented):
the pattern you chose seems like the best choice.
Feels great in the UI.
One possible issue about the tag/meta options available in the dropdown needing to be scoped to current bucket.

frontend/src/components/object/ObjectFilters.vue
```diff
@@ -0,0 +1,138 @@
...
+  const metaToSearch: Array<MetadataPair> = selectedMetadata.value.map(
+    ({ ...meta }: any) => meta
+  );
+  const tagSetToSearch: Array<Tag> = selectedTags.value.map(
+    ({ ...tag }: any) => tag
+  );
+
+  // Search the object store with the tagset as a param and metadata as headers
+  objectStore.fetchObjects(
+    {
```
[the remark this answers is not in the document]
jujaga: I'm 50/50 on this, as while our instance of BCBox will have that on, not all instances of BCBox and COMS will have that as a given. It may be more pedantic, but it might be worth having it be explicitly defined here instead as then there is no room for ambiguity.

frontend/src/components/object/ObjectFilters.vue
jujaga: I agree that we need to add `bucketId` scoping capability to that `/object/tagging` and `/object/metadata` endpoint so that we only get results relevant to this specific bucket, given the context of the page we're doing this on.

frontend/src/components/object/ObjectFilters.vue
```diff
@@ -0,0 +1,138 @@
...
+    @update:model-value="selectedFilterValuesChanged"
+  />
+
+  
+  <MultiSelect
+    v-model="selectedTags"
+    :options="tagsetValues"
+    :loading="searching"
+    class="filterDropdown"
+    display="chip"
```
jujaga, line 122: Screenshot 2023-07-13 at 13-26-49 BCBox - My Objects
The chip is a bit large vertically - if there's a compact option available that would be ideal so that it doesn't touch/spill over the bottom field line.

frontend/src/types/MetadataPair.ts
```diff
@@ -0,0 +1,4 @@
+export type MetadataPair = {
```
jujaga, line 1: It might be better to write this as a generic `KeyValuePair` type instead of domain binding it to metadata, as k/v can be reused later for tag as well for example. It might also be coerceable into a typescript Record<string, string> type, but not quite as familiar with this.

jujaga (commented):
Looking decently nicely on a functional level - we still need to figure out bucketId scoping and other edge case scenarios, but on the high level this looks good.

frontend/src/types/MetadataPair.ts
loneil: I made `MetadataPair` because there was an existing `Tagging` type that used another `Tag` type which is a key value pair, and that was used throughout the application already.
So It was seeming wierd having one of them as a `Record<string, string>` and one a named `Tag` type.
I kind of agree that the generic record one makes more sense but didn't want to go through adjusting existing uses of `Tag`

frontend/src/components/object/ObjectFilters.vue
loneil: Good catch, had to make the dropdown thinner to line up with the search box. Fine tuned the specific sizing for this specific set of multiselect chips

frontend/src/components/object/ObjectFilters.vue
loneil: Lets do like @TimCsaky says for now in here (just added code to filter) where we just filter out items from the metadata and tagging search results to exclude objectIDs that are not in the current list. This takes care of the deleted item issue as well.

I am of the opinion as well that the COMS search endpoints should probably provide bucket (and deleted) filters at the API level. But for now it's simple for BCBox as a consumer at least since (as Tim mentioned) the list of objects is already loaded there.

frontend/src/store/metadataStore.ts
```diff
@@ -89,6 +92,20 @@ export const useMetadataStore = defineStore('metadata', () => {
...
+    metadataSet: Array<MetadataPair> = [],
+  ) {
+    try {
+      state.metadataSearchResults.value = [];
+      // await new Promise((resolve) => setTimeout(resolve, 4000));
+      const response = (await objectService.searchMetadata({ metadata: metadataSet })).data;
+      state.metadataSearchResults.value = response;
+    }
+    catch (error: any) {
+      toast.error('Searching tags', error);
```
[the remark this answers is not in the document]
loneil: Thanks

frontend/src/components/object/ObjectFilters.vue
```diff
@@ -0,0 +1,173 @@
...
+    .map((val) => ({ ...val, display: `${val.key}=${val.value}` }))
+    // coms-id not allowed as a tag to filter on by COMS
+    .filter((val) => val.key !== 'coms-id')
+    // Unique by display property
+    .filter(
+      (val, index, self) =>
+        index === self.findIndex((t) => t.display === val.display)
+    )
+    .sort((a, b) => a.display.localeCompare(b.display));
+});
```
loneil, line 92: This is very duplicate but my attempts to DRY it kept getting difficulties with TypeScript. If anyone has any good ideas I'm all ears

frontend/src/store/objectStore.ts
```diff
@@ -143,9 +158,14 @@ export const useObjectStore = defineStore('object', () => {
 
           // Merge and assign
           state.objects.value = difference.concat(response);
+          // Track all the object IDs that the user would have access to in the table (even if filters are applied)
+          if(!tagset && !metadata) {
```
TimCsaky, line 162: wondering if this needs to check array length = 0. i think the call to fetchObjects from the  filters component always passes in an empty array
eg: `if((!tagset || tagset.length > 0) && (!metadata || metadata.length > 0) {`

frontend/src/components/object/ObjectFilters.vue
```diff
@@ -0,0 +1,171 @@
...
+    v-model="selectedMetadata"
+    :options="metadataValues"
+    :loading="searching"
+    class="filterDropdown mr-4"
+    display="chip"
+    option-label="display"
+    placeholder="Metadata"
+    filter
+    filter-placeholder="Search metadata"
+    :empty-message="searching ? 'Loading...' : 'No available options'"
```
TimCsaky, line 138: i wonder if the 'Loading..' text should only show when the list is empty. the flickering is a bit off-putting.
maybe by setting this to `:empty-message="searching && metadataValues.length ===0 ? 'Loading...' : 'No available options'"`

frontend/src/store/objectStore.ts
```diff
@@ -123,15 +129,24 @@ export const useObjectStore = defineStore('object', () => {
...
+          const headers: Record<string, string> = {};
+          if (metadata?.length) {
+            for (const meta of metadata) {
+              headers[`x-amz-meta-${meta.key}`] = meta.value;
+            }
+          }
+
           response = (await objectService.searchObjects({
             bucketId: params.bucketId ? [params.bucketId] : undefined,
             objectId: uniqueIds,
```
TimCsaky, line 142: not from this PR but i'm not sure this fetahObjects() process is feasible.
the list of objectId's could be way too long to fit in a url, no?

frontend/src/store/objectStore.ts
```diff
@@ -123,15 +129,24 @@ export const useObjectStore = defineStore('object', () => {
...
 
             // Added to allow deletion of objects before versioning implementation
             // TODO: Verify if needed after versioning implemented
             deleteMarker: false,
             latest: true
-          })).data;
+          }, headers)).data;
 
           // Remove old values matching search parameters
           const matches = (x: COMSObject) => (
```
TimCsaky, line 152: regarding the issue where the search results are doing an OR instead of the desired AND behavior, i wonder if we need to remove old values here

TimCsaky (commented):
looking good.
only thing that concerns me is the AND / OR behavior of the search results

frontend/src/components/object/ObjectFilters.vue
TimCsaky: oh and not an issue for me but if we wanted to make the multiselect search field just have the bottom border (as in mockup).. i think the `.p-multiselect-filter .p-inputtext` element can be messed around with.. but not sure if there's a clean way .. or if we should

frontend/src/components/object/ObjectFilters.vue
```diff
@@ -134,5 +163,9 @@ const searchTagging = async () => {
 .filterDropdown {
   min-width: 15rem;
   height: 2.45rem;
+  :deep(.p-multiselect-token) {
+    padding-bottom: 2px;
+    padding-top: 2px;
+  }
```
jujaga, line 169: Not completely sure, but it looks like the words Metadata and Tag in the empty fields are slightly below center field/div with this change.
Screenshot 2023-07-18 at 12-02-14 BCBox - My Objects

frontend/src/store/objectStore.ts
loneil: Yeah it's always put every object ID in there to get the table from the permissions from what I've ever seen. Not sure if that's a sustainable solution yeah.
COMS searching on permissions and doing that in SQL could be a better option 🤷

frontend/src/components/object/ObjectFilters.vue
```diff
@@ -21,22 +22,43 @@ const metadataStore = useMetadataStore();
...
 const { getTagSearchResults } = storeToRefs(useTagStore());
 
 // State
 const searching = ref(false);
 const selectedMetadata = ref([]);
 const selectedTags = ref([]);
 
+// Store subscriptions
+objectStore.$onAction(
+  ({name, args}) => {
```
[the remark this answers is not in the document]
loneil: onAction should always have that
https://pinia.vuejs.org/core-concepts/actions.html#subscribing-to-actions

frontend/src/components/object/ObjectFilters.vue
loneil: Yeah, hadn't seen on my zoom level but at 100% they're a bit off, will tweak

frontend/src/store/objectStore.ts
loneil: Changing to `if(!tagset?.length && !metadata?.length)` to be safe

frontend/src/components/object/ObjectFilters.vue
loneil: The flicker was because before anything was calculated on showing the dropdown the current v-model state would render. Switching to `before-show` is the key to that fix

frontend/src/store/objectStore.ts
```diff
@@ -158,9 +158,13 @@ export const useObjectStore = defineStore('object', () => {
...
-          // Track all the object IDs that the user would have access to in the table (even if filters are applied)
-          if(!tagset && !metadata) {
-            state.unfilteredObjectIds.value = state.objects.value.map((o) => o.id);
+          
+          // Track all the object IDs in this bucket that the user would have access to in the table 
+          // (even if filters are applied)
+          if(!tagset?.length && !metadata?.length) {
+            state.unfilteredObjectIds.value = state.objects.value
+              .filter( (x) => x.bucketId === params.bucketId )
+              .map((o) => o.id);
```
jujaga: This works for now, but if and when we decide to revisit the idea of aggregating all objects across all buckets a user has access to, this new logic will be in conflict with this. Something we'll have to assess down the line if we decide to implement a cross-bucket view like that.

jujaga (approved):
Tentatively approving as most major aspects appear to be functioning as intended now.

TimCsaky (approved):

frontend/src/store/objectStore.ts
loneil: Yeah this will need to be adjusted if it's across buckets when/if that becomes a feature

frontend/src/components/object/ObjectFilters.vue
```diff
@@ -88,6 +95,29 @@ const tagsetValues = computed(() => {
 });
 
 // Actions
+const uncheckOther= (event: MultiSelectChangeEvent, modelValue: Ref<MetadataPair[] | Tag[]>) =>{
```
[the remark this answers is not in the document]
loneil: fixed

frontend/src/components/object/ObjectFilters.vue
```diff
@@ -1,12 +1,19 @@
 <script setup lang="ts">
 import { storeToRefs } from 'pinia';
-import { ref, computed } from 'vue';
+import { ref, computed, type Ref } from 'vue';
```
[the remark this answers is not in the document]
loneil: fixed

Issue #100, opened 2023-06-28
Restrict sharing files publicly


danebenal:
#### Is your feature request related to a problem? Please describe.
There are concerns with how easy it is to make files publicly accessible. It is currently just one click to make a file publicly accessible, and some users might click the button inadvertently. Even having a confirmation box might not be enough as some users are trained to click through dialog boxes without fully reading them.

#### Describe the solution you'd like
Make sharing files publicly more difficult:
- a setting at the bucket level to allow/disallow files being shared publicly; or
- new user permission to be able to set files to public

slhurley:
Right now the MANAGE permission handles this.  You must have MANAGE on the bucket or MANAGE on the object to use the public toggle.  I think users need to ensure they are handing out those privileges carefully.  Of course anyone who "owns" a file by being the one who did the upload in the first place will have MANAGE on that file as it is theirs to grant further permission as they see fit.

danebenal:
There are cases where none of the files within a bucket would be made public and having the toggle so easily accessible (and just one click) can lead to inadvertently making files public, potentially causing a privacy breach if PI data is being transferred.

Someone uploading a file does not necessarily mean they "own" the file or content -- they may be under contract to provide the file and are using this system to securely transfer it back to Government users.

TimCsaky:
Done. Thanks Dan

jujaga:
In order to use the public toggle, **both** the following criteria must be satisfied:

1) You are logged in as a user with elevated rights (IDIR)
2) You must have `MANAGE` on the file in question

Otherwise, it will be disabled and grayed out.

PR #104, opened 2023-07-18
Disable "public" toggle for files if not elevated user

norrisng-bc:
# Description

Disables the "public" toggle for files if the user isn't an elevated (i.e. IDIR) user.

[SHOWCASE-3281](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3281)

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/object/ObjectTable.vue
+++ b/frontend/src/components/object/ObjectTable.vue
@@ -202,9 +202,10 @@ const filters = ref({
         <template #body="{ data }">
           <InputSwitch
             v-model="data.public"
-            :disabled="!permissionStore.isObjectActionAllowed(
-              data.id, getUserId, Permissions.MANAGE, props.bucketId as string) ||
-              !usePermissionStore().isUserElevatedRights()"
+            :disabled="!(
+              usePermissionStore().isUserElevatedRights() &&
+              permissionStore.isObjectActionAllowed(
+                data.id, getUserId, Permissions.MANAGE, props.bucketId as string))"
             @change="togglePublic(data.id, data.public)"
           />
         </template>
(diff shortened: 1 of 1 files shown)


frontend/src/components/object/ObjectTable.vue
```diff
@@ -203,7 +203,8 @@ const filters = ref({
           <InputSwitch
             v-model="data.public"
             :disabled="!permissionStore.isObjectActionAllowed(
-              data.id, getUserId, Permissions.MANAGE, props.bucketId as string)"
+              data.id, getUserId, Permissions.MANAGE, props.bucketId as string) ||
+              !usePermissionStore().isUserElevatedRights()"
```
jujaga, line 207: I'm half tempted to recommend applying the De Morgan's Law inverse to this boolean statement in order to improve readability, as this has a lot of negative assertions at the moment. I also suspect that if we evaluate `isUserElevatedRights` first instead of `isObjectActionAllowed`, we can take advantage of short circuiting and potentially reduce one network call.
```ts
:disabled="!(
  usePermissionStore().isUserElevatedRights() && 
  permissionStore.isObjectActionAllowed(
    data.id, getUserId, Permissions.MANAGE, props.bucketId as string))"
```

norrisng-bc: I think what makes it so confusing / hard to read is that the Vue prop itself being a negative itself means that we need some level of double negation in there - i.e. you *have* to tell Vue "do **not** enable this if...", instead of "**only** enable if...".

But I agree with you on all counts there.

jujaga (approved):

TimCsaky (approved):
just as confusing to me

TimCsaky:
addresses issue #100

PR #107, opened 2023-07-22
Test binary

loneil:
This is just playing around with uploading a binary with axios from the BCBox Frontend to the object store URL.

For this example it's taking the file you provide and using application/octet-stream instead of multipart to PUT it over top of an existing file url, creating a new version.
Of course this doesn't call COMS so nothing updates in the database in this POC, but I can see in S3 browser the replacement happening. Basically I set this up in Postman and copeied that flow over to BCBox.

I have tried this only with simple small .txt files in this case.

So what is needed for this to work in BCBox in my limited test is:

**1 CORS allowance in the bucket.** 

I didn't leave anything set don't want to mess with `egejyy` bucket perms, but this CORS setting should work. Might want to tweak for specific origins or something
```
<CORSConfiguration>
  <CORSRule>
    <AllowedOrigin>*</AllowedOrigin>
    <AllowedMethod>PUT</AllowedMethod>
    <AllowedHeader>*</AllowedHeader>
  </CORSRule>
</CORSConfiguration>
```

Looks like you can set it in S3 browser

image

image

Not sure how to consider this on a Bring Your Own Bucket model...
Looks like common services bucket doesn't really have any production data so maybe you can leave a permissive setting on while trying things out.
(there are CHEFS objects still in there from a year ago FYI)

**2 The call from the FE needs the signature header.** 

I could only figure out how to do this in Postman, but there's a handy tool in there for getting AWS headers from credentials. See below

image

image

So to test, can use the Postman thing to get the generated headers and then I added input boxes in BCBox to paste in there.
(If you toggle back and forth between the tabs and alter something it should regen the new headers each time, there's a time limit on the auth header I believe so can't use it for too long)

Obviously if calling from BCBox browser to object store URL is the way you want to go you can't have the secret there in the browser to generate the signature so will have to figure out something for that.

So if you do the above, you should be able to do a file binary PUT from the FE to a existing file url and it will write a new version.

Hit the Upload button on BCBox to reveal the POC code. It has the instructions for what you can try printed there.

image

None of it is really wired up to Vue events or anything, so once you Choose File it will run it, monitor network tab to see what happens.
Failures include CORS Error, or 403 if signature or link or something is invalid. Like I said, really only tried with small .txts for this.

Hope this helps show something! Not particularly deep.
I'm off out of service after this but will be back on the 27th/28th if there's any questions.

### DIFF
--- a/frontend/src/components/object/ObjectUpload.vue
+++ b/frontend/src/components/object/ObjectUpload.vue
@@ -3,7 +3,7 @@ import { storeToRefs } from 'pinia';
 import { ref } from 'vue';
 
 import ObjectUploadFile from '@/components/object/ObjectUploadFile.vue';
-import { Button, FileUpload, useToast } from '@/lib/primevue';
+import { Button, FileUpload, InputText, useToast } from '@/lib/primevue';
 import { useAuthStore, useAppStore, useObjectStore } from '@/store';
 
 import type { Ref } from 'vue';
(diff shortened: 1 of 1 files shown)


jujaga:
Superceded by #111

PR #111, opened 2023-08-15
SC-3303: Align BCBox with V0.6 file upload changes

wilwong89:
# Description
Update BCBox to use new API for COMS.
Change createObject and updateObject methods to use PUT instead of POST
Upload as octet-stream

## Types of changes

Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [X] I have added necessary documentation (if appropriate)

## Further comments
Made an assumption that objects passed to objService will be of `File` type.

### DIFF
--- a/frontend/src/services/objectService.ts
+++ b/frontend/src/services/objectService.ts
@@ -290,12 +290,6 @@ export default {
       },
     };
 
-    // Detect MIME type from File object and assign to header
-    // Fall back to application/octet-stream as above
-    if (object?.type) {
-      config.headers['Content-Type'] = object.type;
-    }
-
     // Map the metadata if required
     if (headers.metadata) {
       config.headers = {
--- a/frontend/src/services/objectService.ts
+++ b/frontend/src/services/objectService.ts
@@ -28,8 +28,9 @@ export default {
   ) {
     const config = {
       headers: {
-        'Content-Type': 'application/octet-stream',
         'Content-Disposition': setDispositionHeader(object.name)
+        // Content-Type header/MIME type detection needs TBD
+        // Also in uploadObject call below
       },
       params: {
         bucketId: params.bucketId,
@@ -278,8 +279,8 @@ export default {
   ) {
     const config = {
       headers: {
-        'Content-Type': 'application/octet-stream',
         'Content-Disposition': setDispositionHeader(object.name)
+        // Content-Type header/MIME type detection needs TBD
       },
       params: {
         tagset: {}
--- a/frontend/src/services/objectService.ts
+++ b/frontend/src/services/objectService.ts
@@ -1,4 +1,5 @@
 import { comsAxios } from './interceptors';
+import { setDispositionHeader } from '@/utils/utils';
 
 import type { AxiosRequestConfig } from 'axios';
 import type { GetMetadataOptions, GetObjectTaggingOptions, MetadataPair, SearchObjectsOptions, Tag } from '@/types';
@@ -14,7 +15,7 @@ export default {
    * @param {AxiosRequestConfig} axiosOptions Axios request config options
    * @returns {Promise} An axios response
    */
-  createObject(
+  async createObject(
     object: any,
     headers: {
       metadata?: Array<{ key: string; value: string }>,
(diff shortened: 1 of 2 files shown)


frontend/src/utils/utils.ts
```diff
@@ -62,3 +62,20 @@ export function partition<T>(
     [[], []] as [Array<T>, Array<T>]
   );
 }
+
+/**
+ * @function setDispositionHeader
+ * Constructs a value for http request header 'Content-Disposition'
```
jujaga, line 68: Constructs a valid RFC 6266 'Content-Disposition' header and optionally handles RFC 8187 UTF-8 encoding when necessary

frontend/src/services/objectService.ts
```diff
@@ -295,8 +301,7 @@ export default {
       );
     }
 
-    const fd = new FormData();
-    fd.append('file', object);
-    return comsAxios(axiosOptions).post(`${PATH}/${objectId}`, fd, config);
+    const fd = await object.arrayBuffer();
```
jujaga, line 304: I'm not completely certain what the await does, but does this end up forcing the browser to consume a large file payload (multiple gigs) and shove into memory if we end up awaiting? Just slightly concerned about potential adverse effects to browser memory usage on significantly large arrayBuffered files.

frontend/src/services/objectService.ts
```diff
@@ -26,7 +27,10 @@ export default {
     axiosOptions?: AxiosRequestConfig
   ) {
     const config = {
-      headers: { 'Content-Type': 'multipart/form-data' },
+      headers: {
+        'Content-Type': 'application/octet-stream',
+        'Content-Disposition': setDispositionHeader(object.name)
+      },
```
jujaga, line 33: Two things:
1. Please alphabetize the headers - Disposition should come before Type
2. We default to mimetype `application/octet-stream`, but we can't say every file is of this type. Please take a look at the `file-type` npm package or equivalent to have it inspect the incoming arrayBuffer stream and determine what kind of mimetype it is.

frontend/src/services/objectService.ts
```diff
@@ -274,7 +277,10 @@ export default {
...
```
jujaga, line 283: Two things:
1. Please alphabetize the headers - Disposition should come before Type
2. We default to mimetype `application/octet-stream`, but we can't say every file is of this type. Please take a look at the `file-type` npm package or equivalent to have it inspect the incoming arrayBuffer stream and determine what kind of mimetype it is.

jujaga (commented):
Definitely glad this pathway actually works with axios without too much relative headache. That being said, we need to make sure we're actually doing some proper handling of the mimetype detection and not just calling everything an octet-stream.

frontend/src/utils/utils.ts
wilwong89: Added

frontend/src/services/objectService.ts
wilwong89: Content type removed, will resolve in another ticket.

frontend/src/services/objectService.ts
wilwong89: After testing on some images, it doesn't seem like there's a synchronous lock on the browser while it uploads. But that isn't a full test on browser memory and possible adverse effects.

jujaga (approved):

norrisng-bc (approved):

frontend/src/services/objectService.ts
```diff
@@ -28,16 +28,21 @@ export default {
...
         tagset: {}
       },
     };
 
+    // Detect MIME type from File object and assign to header
+    // Fall back to application/octet-stream as above
+    if (object?.type) {
+      config.headers['Content-Type'] = object.type;
+    }
+
```
jujaga, line 45: We can collapse the mimetype logic down into an in-stream one-liner
```ts
headers: {
  'Content-Disposition': setDispositionHeader(object.name),
  'Content-Type': object?.type ? object.type : 'application/octet-stream'
},
```

frontend/src/services/objectService.ts
```diff
@@ -279,14 +284,20 @@ export default {
...
```
jujaga, line 300: We can collapse the mimetype logic down into an in-stream one-liner
```ts
headers: {
  'Content-Disposition': setDispositionHeader(object.name),
  'Content-Type': object?.type ? object.type : 'application/octet-stream'
},
```

PR #116, opened 2023-08-23
Divide COMS object search request to limit URL length

TimCsaky:
# Description

ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3330

to load the objects table, if there are more than about 50 objects some browsers will show errors, because the COMS request url contains so many `objectId`'s and is too long.

In object search service function, i have split into multiple requests, each time, passing an acceptable number of the `objectId`'s

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/services/objectService.ts
+++ b/frontend/src/services/objectService.ts
@@ -1,6 +1,6 @@
 import { comsAxios } from './interceptors';
 import { setDispositionHeader } from '@/utils/utils';
-import { ConfigService } from './index';
+import ConfigService from './configService';
 
 import type { AxiosRequestConfig } from 'axios';
 import type { GetMetadataOptions, GetObjectTaggingOptions, MetadataPair, SearchObjectsOptions, Tag } from '@/types';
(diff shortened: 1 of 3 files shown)


frontend/tests/unit/store/objectStore.spec.ts
```diff
@@ -181,7 +181,7 @@ describe('Object Store', () => {
         readPerm
       ];
 
-      searchObjectsSpy.mockReturnValue({ data: [obj] } as any);
+      searchObjectsSpy.mockReturnValue(Promise.resolve({ data: [obj] }) as any);
```
jujaga, line 184:
```ts
searchObjectsSpy.mockResolvedValue({ data: [obj] } as any);
```

frontend/src/services/objectService.ts
```diff
@@ -227,7 +228,54 @@ export default {
...
+
+    if (params.objectId) {
+      /**
+       * split calls to COMS if query params (eg objectId's)
+       * will cause url length to excede 2000 characters
+       * see: https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers
+       *
+       * TODO: consider creating a utils function
+       * eg: `divideParam(params, attr)`
+       *     ... return Promise.all(divideParam(params, objectId).map(zparam => comsAxios().get(PATH, {params: zparam, headers: headers});
```
jujaga, line 240: Break up the following comment line - it's just too long even for our permissive linter.
> This line has a comment length of 138. Maximum allowed is 120

frontend/src/services/objectService.ts
```diff
@@ -227,7 +228,54 @@ export default {
...
+      urlLimit = urlLimit - 19;
+      // minus latest=false
+      urlLimit = urlLimit - 13;
+      // minus a single bucketId (bucketId[]=<uuid>)
+      if (params.bucketId) { urlLimit = urlLimit - 48; }
+      // if tagset parameters passed
+      if (params.tagset) {
+        for (const [key, value] of Object.entries(params.tagset)) {
+          // @ts-ignore
+          urlLimit = urlLimit - 10 - key.length - value.length;
```
jujaga, line 256: I'm assuming you're encountering some shenanigans with typescript to just put in a blanket ignore? We should try to resolve the problem instead of just ignoring it, even if it is a bit pedantic. In the event it's the lack of definition for `Object.prototype.entries`, we could try this typing:
```ts
for (const [key, value] of Object.entries<string>(params.tagset): {[K in keyof T]: [K, T[K]]}[keyof T][]) {
  urlLimit -= 10 + key.length + value.length; // subtract k/v pair and boilerplate
}
```
Some inspiration from here: <https://github.com/microsoft/TypeScript/issues/35101>
If it doesn't work, it's still better we force-cast it as an `any` type instead of ignoring it if possible.

frontend/src/services/objectService.ts
```diff
@@ -227,7 +228,54 @@ export default {
...
+      let urlLimit = 2000;
+      // minus base url
+      const baseUrl = new URL(`${new ConfigService().getConfig().coms.apiPath}${PATH}`).toString().length;
+      urlLimit = urlLimit - baseUrl;
+      // minus deleteMarker=false
+      urlLimit = urlLimit - 19;
+      // minus latest=false
+      urlLimit = urlLimit - 13;
+      // minus a single bucketId (bucketId[]=<uuid>)
+      if (params.bucketId) { urlLimit = urlLimit - 48; }
```
jujaga, line 251: - May be worth rewriting these using subtraction assignment instead as it'll make it more concise.
- The const assignment on Line 244 should be the baseUrl string, not the number value thereof.
- The conditional body on Line 251 doesn't need curly brace scoping as it's just a one-liner.

```ts
const baseUrl = new URL(`${new ConfigService().getConfig().coms.apiPath}${PATH}`).toString();
urlLimit -= baseUrl.length; // subtract baseUrl length
urlLimit -= 19; // subtract `deleteMarker=false`
urlLimit -= 13; // subtract `latest=false`
if (params.bucketId) urlLimit -= 48; // subtract a single bucketId `bucketId[]=<uuid>`
```

wilwong89: Will parameters like `deleteMarker=false` and `latest=false` be guaranteed to exist? Or should some logic be used to check for them before subtracting the limit as future proofing?

jujaga (approved):

wilwong89 (approved):

PR #121, opened 2023-09-12
SC-3258, feature: Add sync button+dialog for both bucket+object respe…

wilwong89:
# Description

Add sync button to bucket list page for each bucket.
That button will trigger a dialog box to start bucket sync.

Add sync button to object list page for each object.
That button will trigger a dialog box to start object sync.

## Types of changes

## Checklist

## Further comments

Needs Tyler to check if UI is to spec. Screenshots included in https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3258
Needs someone to double check sync calls and syncing feature works as expected.


jujaga (commented):
Visually seems correct. A few comments on code structure however:
- We should not be in-lining the syncBucket dialog and action behavior as a part of the BucketList component; this should be a standalone component at minimum similar to your new SyncObjectButton.
- Followup from above: we should unify both the SyncObjectButton and the new SyncBucket equivalent to both point to one component called "SyncButton" which resides under root `/components` folder. This component should take in two optional parameters; either an objectId, or a bucketId; depending on which was passed in, it will know to call either syncObject or syncBucket store methods appropriately.

frontend/src/components/common/SyncButton.vue
```diff
@@ -1,38 +1,48 @@
...
+const bucketStore = useBucketStore();
+const toast = useToast();
 
 // Dialog
 const displaySyncDialog = ref(false);
 
 // Actions
-const syncObject = () => {
-  objectStore.syncObject(props.id);
+const syncFunction = () => {
```
jujaga, line 30: Might be better to just call this onSubmit instead

frontend/src/components/common/SyncButton.vue
```diff
@@ -1,38 +1,48 @@
...
 
-import type { Ref } from 'vue';
-import type { COMSObject } from '@/types';
 
 // Props
 type Props = {
+  type: string;
   id: string;
+  name: string;
 };
```
jujaga, line 14: Is there a reason why we didn't do props like this instead?
```ts
type Props = {
  bucketId?: string,
  objectId?: string
};
```
Forcing a name dependency upstream as a required prop doesn't appear to be a great idea; the component should be the one being able to figure out that information itself. Also if we want to be doing an explicit type string instead, this should be defined as an enum type, not as two constant strings.

frontend/src/components/object/ObjectTable.vue
```diff
@@ -9,7 +9,8 @@ import {
   ObjectFilters,
   ObjectPermission,
 } from '@/components/object';
-import { ShareObjectButton, SyncObjectButton } from '@/components/object/share';
+import { ShareObjectButton } from '@/components/object/share';
+import { SyncButton } from '@/components/common';
```
jujaga, line 13: Switch the import ordering - we sort based on the from string, not the component name.

frontend/src/components/common/SyncButton.vue
```diff
@@ -0,0 +1,115 @@
+<script setup lang="ts">
+import { ref } from 'vue';
+
+import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
+import { Button, Dialog, useToast } from '@/lib/primevue';
+import { useObjectStore, useBucketStore } from '@/store';
+
+
```
jujaga, line 8: Extra unnecessary whitespace line

frontend/src/store/bucketStore.ts
```diff
@@ -118,7 +133,8 @@ export const useBucketStore = defineStore('bucket', () => {
     deleteBucket,
     fetchBuckets,
     findBucketById,
-    updateBucket
+    updateBucket,
+    syncBucket
```
jujaga, line 137: This should be in alphabetical order.

frontend/src/store/objectStore.ts
```diff
@@ -266,7 +280,8 @@ export const useObjectStore = defineStore('object', () => {
     headObject,
     setSelectedObjects,
     togglePublic,
-    updateObject
+    updateObject,
+    syncObject
```
jujaga, line 284: Alphabetical order

frontend/src/components/common/SyncButton.vue
```diff
@@ -0,0 +1,115 @@
...
+    <font-awesome-icon icon="fa-solid fa-sync" />
+  </Button>
+</template>
+
+<style scoped lang="scss">
+h2 {
+  font-weight: bold;
+}
+ul {
+  padding-left: 22px;
```
jujaga, line 113: Unless this was explicitly defined, it may be better to have this be defined in either rem or em units instead of explicit pixels to maintain relativity calculations.

jujaga (approved):

TimCsaky (approved):
good work!

norrisng-bc (approved):

PR #134, opened 2023-10-16
feat:(SC-3320) Added tests for layout components and bucket components

wilwong89:
# Description
For ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3320
Added a few frontend component tests. 
Test for layout components:
-Footer
-Navbar
-ProgressLoader
-Spinner

Tests for bucket components:
-BucketConfigForm
-BucketList
-BucketPermission
-BucketPermissionAddUser
-BucketSidebar
-BucketTable

Not all tests are fully complete for the above files, will need more work to improve coverage.

## Types of changes

## Checklist

- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/frontend/tests/unit/components/bucket/BucketConfigForm.spec.ts
+++ b/frontend/tests/unit/components/bucket/BucketConfigForm.spec.ts
@@ -1,0 +1,79 @@
+import { createTestingPinia } from '@pinia/testing';
+import { mount, shallowMount, RouterLinkStub } from '@vue/test-utils';
+
+import BucketConfigForm from '@/components/bucket/BucketConfigForm.vue';
+import * as primevue from '@/lib/primevue';
+import { StorageKey } from '@/utils/constants';
+import PrimeVue from 'primevue/config';
+import ToastService from 'primevue/toastservice';
+
+const mockToast = vi.fn();
+const useToastSpy = vi.spyOn(primevue, 'useToast');
+
+beforeEach(() => {
+  sessionStorage.setItem(StorageKey.CONFIG, JSON.stringify(
+    {
+      oidc: {
+        authority: 'abc',
+        clientId: '123'
+      }
+    }
+  ));
+
+  vi.clearAllMocks();
+  useToastSpy.mockImplementation(() => ({ error: mockToast, info: mockToast, success: mockToast, warn: mockToast }));
+});
+
+afterEach(() => {
+  sessionStorage.clear();
+});
+
+describe('BucketConfigForm.vue', () => {
+  it('renders', () => {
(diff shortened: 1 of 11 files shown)


frontend/tests/unit/components/bucket/BucketConfigForm.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+        stubs: {
+          RouterLink: RouterLinkStub
+        },
+      },
+      props: {}
+    });
+    expect(wrapper).toBeTruthy();
+  });
+
+  it('it emits cancel config', async () => {
```
kyle1morel, line 45: double 'it'

frontend/tests/unit/components/bucket/BucketConfigForm.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+      props: {}
+    });
+
+    const cancelButton = wrapper.get('[aria-label="Cancel"]');
+    await cancelButton.trigger('click');
+
+    expect(wrapper.emitted()).toHaveProperty('cancel-bucket-config');
+  });
+
+  it('it emits triggers onSubmit error', async () => {
```
kyle1morel, line 62: double 'it'

frontend/tests/unit/components/bucket/BucketList.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+          pinia,
+        ],
+        stubs: [
+          'font-awesome-icon',
+          'BucketTable'
+        ]
+      },
+    });
+
+    const button = wrapper.find('[icon="fa-solid fa-plus"]');
```
kyle1morel, line 76: Wondering if we should be attaching proper IDs to these instead of searching by their icons.

jujaga (approved):

frontend/tests/unit/components/bucket/BucketList.spec.ts
jujaga: Buttons should always have some form of unique key or identifier. In the past, we used to insert `data-test` attribute manually into various DOM elements so that we could succinctly select and then test on them later. This might not be as useful at the unit test level, but becomes critically important if and when we ever get to integration or end to end testing level using tools such as cypress.

wilwong89: I think proceeding with adding `data-test` attributes is a good idea in the future. Perhaps future test writers should add them in as necessary when writing more tests.

TimCsaky (approved):

kyle1morel (approved):

PR #135, opened 2023-10-16
Sc3004

jatindersingh93:
# Description
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3004
- Added aria-label attributes to links/buttons that open in new tab or popup
- Restructure Heading tags in template
- Included alt text for images

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/bucket/BucketList.vue
+++ b/frontend/src/components/bucket/BucketList.vue
@@ -64,6 +64,7 @@ onMounted(async () => {
         class="p-button-outlined mt-4"
         data-test="connect-bucket"
         @click="showBucketConfig()"
+        aria-label="Configure bucket"
       >
         <font-awesome-icon icon="fa-solid fa-plus" />
         Connect bucket to BCBox
--- a/frontend/src/components/bucket/BucketList.vue
+++ b/frontend/src/components/bucket/BucketList.vue
@@ -48,8 +48,8 @@ onMounted(async () => {
 <template>
   <div>
     <div>
-      <h2>Select a bucket</h2>
-      <h3>Buckets are containers for storing objects.</h3>
+      <h1>Select a bucket</h1>
+      <h2>Buckets are containers for storing objects.</h2>
       <Message
         v-if="getConfig?.notificationBanner"
         severity="warn"
--- a/frontend/src/components/bucket/BucketList.vue
+++ b/frontend/src/components/bucket/BucketList.vue
@@ -48,7 +48,7 @@ onMounted(async () => {
 <template>
   <div>
     <div>
-      <h1>Select a bucket</h1>
+      <h2>Select a bucket</h2>
       <h3>Buckets are containers for storing objects.</h3>
       <Message
         v-if="getConfig?.notificationBanner"
--- a/frontend/src/components/bucket/BucketList.vue
+++ b/frontend/src/components/bucket/BucketList.vue
@@ -63,8 +63,8 @@ onMounted(async () => {
         label="Primary"
         class="p-button-outlined mt-4"
         data-test="connect-bucket"
-        @click="showBucketConfig()"
         aria-label="Configure bucket"
+        @click="showBucketConfig()"
       >
         <font-awesome-icon icon="fa-solid fa-plus" />
         Connect bucket to BCBox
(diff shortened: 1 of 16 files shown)


wilwong89:
There's a whole bunch of linting warnings about `aria-label` and line length. I think it would be a good idea to resolve those. They show up in the "files changed" tab on this page. Otherwise it looks good.

frontend/src/components/bucket/BucketList.vue
```diff
@@ -48,7 +48,7 @@ onMounted(async () => {
 <template>
   <div>
     <div>
-      <h1>Select a bucket</h1>
+      <h2>Select a bucket</h2>
```
kyle1morel, line 51: As this is the main page header this should still be an h1 according to the requirements.

> Requirement: An H1 title provides information to blind-users using screen-readers of what the main topic of the page is and each page should have exactly one H1 title.

jatindersingh93: 'BCBox' is already set to H1 in the Header, I am not sure what the intention was but looks like it serve the purpose and verify the statement. I didn't change the structure considering it has been verified and accepted during initial application acceptance or accessibility testing. suggestion?

frontend/src/components/bucket/BucketTable.vue
jujaga, line 1: General capitalization of the aria-label contents should be standardized as they seem inconsistent; some are caps and lowercase, others are caps all the way through.

frontend/src/components/bucket/BucketList.vue
TimCsaky: i personally dont think the BCBox in the blue header should be h1. 
I would probably change the header.vue be like this:
`<div
            id="title-branding"
            class="justify-content-left"
            role="banner"
          >
            <span>BCBox</span>
          </div>`
 and do something in the main.scss like:
  `.title-branding > span { @extend h1 }`

frontend/src/components/layout/Navbar.vue
```diff
@@ -25,14 +25,18 @@ const { getIsAuthenticated } = storeToRefs(useAuthStore());
...
+              aria-label="My Buckets"
+            >
               My Buckets
             </router-link>
           </li>
           <li class="mr-2">
             <a
               target="_blank"
               href="https://github.com/bcgov/bcbox/wiki"
+              aria-label="BCBox Wiki"
```
TimCsaky, line 39: Maybe this should be 'BCBox Help' 🤷 but either is fine by me

frontend/src/components/bucket/BucketList.vue
TimCsaky: this would mean changing the main header in each view to h1.

TimCsaky (commented):
looking good

frontend/src/components/bucket/BucketList.vue
tpantella: Hi all. I see the conundrum because the Design System describes it as H1:
https://developer.gov.bc.ca/Design-System/Header-Basic
But if you go to any given gov.bc.ca page, they use an actual proper H1-level heading for unique page topics and the header is blank:
https://www2.gov.bc.ca/gov/content/health/health-drug-coverage/pharmacare-for-bc-residents/pharmacare-information-sheets

gov.bc.ca uses a single H1 as a unique page title (could be interpreted as a single main topic) then H2, H3, H4 as heading levels for topics. I think I should follow up with the Design System folks (even though it's being revamped) because it's inconsistent with standards set by gov.bc.ca and the intention of H1 as described by W3 and in accessibility standards.

With all of the above information, I would suggest that we style the first black-on-white unique heading on each page as H1, such as My Buckets. BCBox header should not be H1 because it's universal, it's redundant to force screen reader accessibility software to describe it as an H1 on each unique page view. I like Tim's idea of changing header.vue and then setting each "unique page title" to H1.

kyle1morel (approved):

frontend/src/components/layout/Navbar.vue
tpantella: I agree, we refer to it as "Help" in the interface so "BCBox Help" gets my vote

TimCsaky (approved):

jujaga (approved):

PR #136, opened 2023-10-17
fix:(SC-3376) Handle delete success/fail toasts on a per object basis

wilwong89:
# Description

objectStore.ts now uses Promise.allSettled instead of Promise.all and returns an array of objects for each promise with their status.

DeleteObjectButton.vue waits for the array and displays a toast based on each success or fail.

## Types of changes

## Checklist

- [] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if appropriate)

## Further comments
Deleting a list of 3 files looks as follows. 

Screenshot 2023-10-17 at 2 28 49 PM


frontend/src/components/object/DeleteObjectButton.vue
```diff
@@ -42,8 +42,15 @@ const confirmDelete = () => {
       rejectLabel: 'Cancel',
       accept: async () => {
         try {
-          await objectStore.deleteObjects(props.ids, props.versionId);
-          emit('on-deleted-success', props.versionId);
+          const res = await objectStore.deleteObjects(props.ids, props.versionId);
+          res?.forEach((ele, index)=>{
```
jujaga, line 46: Formatting/spacing with the arrow.

While this does technically work, this does have the issue of blocking immediacy of responses. While this is a delete so shouldn't be affected, consider if you were instead uploading one small and one very large file. Promise.allSettled will force both alerts. big and small, to stall out and wait for both to be complete before you ever get the alert, even though the small file may have finished a long time ago. It would be more ideal to consider instead, creating an array of singular objectStore.deleteObject calls instead; then the component here would have agency over the entire asynchronous chain of actions on each individual file (and dispatching appropriately) instead of shunting everything into a promiseSettled and making everything have to wait until hard completion.

frontend/src/store/objectStore.ts
```diff
@@ -81,7 +81,7 @@ export const useObjectStore = defineStore('object', () => {
 
     try {
       appStore.beginIndeterminateLoading();
-      await Promise.all(
+      return await Promise.all(
```
jujaga, line 84: What's the return value when we catch/throw? The return here may also interfere with the finally block from executing. As this is a function signature change (this function used to return void, now it's returning a Promise) we need to make sure this is actually working on all pathways.

frontend/src/components/object/DeleteObjectButton.vue
```diff
@@ -37,13 +37,15 @@ const confirmDelete = () => {
...
       acceptLabel: 'Confirm',
       rejectLabel: 'Cancel',
       accept: async () => {
         try {
-          await objectStore.deleteObjects(props.ids, props.versionId);
-          emit('on-deleted-success', props.versionId);
+          props.ids?.forEach(async (ele) => {
+            const res = await objectStore.deleteObjects([ele], props.versionId);
+            if (res) emit('on-deleted-success', props.versionId);
+          });
```
jujaga, line 48: The way this is written, I suspect if there are multiple failures, this will not handle that as the next real catch halts the forEach iteration, so this is not truly asynchronous behavior.

TimCsaky (approved):
This all works well now.
aside from the annoying formatting warnings.

jujaga (approved):

frontend/tests/unit/store/objectStore.spec.ts
```diff
@@ -70,7 +70,7 @@ describe('Object Store', () => {
 
   let createObjectSpy: SpyInstance;
   let deleteObjectSpy: SpyInstance;
-  let fetchObjectsSpy: SpyInstance;
+  // let fetchObjectsSpy: SpyInstance;
```
jatindersingh93, line 73: Not a big concern, just wondering why fetchObjectsSpy are commented and not removed?

jatindersingh93 (approved):

PR #137, opened 2023-10-18
Filter S3 Metadata in COMS data

TimCsaky:
# Description

ticket: https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3379

NOTE: we still need to decide on the actual keys we want to exclude. Maybe there's an easy way to update that in future.

general approach:
- remove metadata from the coms response in bcbox service later (object service.getMetadata())
- excluded metadata doesnt get into bcbox store
- if they edit metadata on an object (using bcbox edit metadata form) the new version will not have the excluded metadata
- coms-id tag will always persist
- removed some old code to filter coms-id during update

to test, add this to your local.json and clear browser session storage(existing config):
```json
{
  "frontend": {
    "exclde": {
      "metadata": "geodrive.common.encoding,geodrive.windows.att,geodrive.windows.secdesc,s3b-last-modified",
      "tagset": "coms-id"
    },
    "oidc": {
```

## Types of changes

## Checklist

## Further comments
- may want to consider hiding ugly meta rather than excluding in a future PR
- i think we're duplicating code with searchMetadata() and getMetadata() in the object service.


frontend/src/utils/utils.ts
```diff
@@ -63,6 +64,27 @@ export function partition<T>(
...
+/**
+ * @function excludeMetaTag
+ * Filter out a configured list of select metadata or tags from a COMS response
+ * @param {object} type either 'metadata' or 'tagset'
+ * @param {object} axiosResponse A resolved axios response from COMS
+ * @returns {object} The response data with select metadata/tags from a configured list removed
+ */
+export function excludeMetaTag(type: string, data: any){
+  // array of selected tags/metadata (keys) to hide from UI
+  const excludeArray = new ConfigService().getConfig().exclude?.[type].split(',') ?? [];
```
jujaga, line 76: Need to do a `trim` after the split as we are not guaranteed that environment variables are well formatted - this will at least handle any errant whitespaces users may put into the environment variable.
```ts
// Array of specific values to mask
const excludeArray = new ConfigService().getConfig().exclude?.[type].split(',').map(s => s.trim()) ?? [];
```

frontend/src/utils/utils.ts
```diff
@@ -63,6 +64,27 @@ export function partition<T>(
...
+  // filter COMS data
+  return data.map((obj: any) => {
+    return {
+      ...obj,
+      [type]: obj[type]?.filter((el: any) => {
+        return !excludeArray.includes(el.key);
+      })
+    };
+  });
```
jujaga, line 85: We can leverage closures and inlining to reduce the amount of explicit return statements needed for the operation.
```ts
// Exclude specific values
return data.map((obj: any) => ({
  ...obj,
  [type]: obj[type]?.filter((el: any) => !excludeArray.includes(el.key))
}));
```

frontend/src/utils/utils.ts
```diff
@@ -63,6 +64,27 @@ export function partition<T>(
...
 }
 
+/**
+ * @function excludeMetaTag
+ * Filter out a configured list of select metadata or tags from a COMS response
+ * @param {object} type either 'metadata' or 'tagset'
+ * @param {object} axiosResponse A resolved axios response from COMS
+ * @returns {object} The response data with select metadata/tags from a configured list removed
+ */
+export function excludeMetaTag(type: string, data: any){
```
jujaga, line 74: - As excludeMetaTag is a new function, don't forget to add unit tests for this utility.
- Wondering if we should be considering using an enum type for this instead of just a string, as the function name implies only metadata or tags, whereas we accept basically any valid string here. This can easily lead to an error (that we do catch with the ?? operator) if we pass in a string that doesn't exist in the exclude object.

frontend/src/services/versionService.ts
```diff
@@ -20,6 +25,10 @@ export default {
    * @returns {Promise} An axios response
    */
   getObjectTagging(params: GetVersionTaggingOptions) {
-    return comsAxios().get(`${PATH}/tagging`, { params: params });
+    return comsAxios().get(`${PATH}/tagging`, { params: params })
+      // filter out a configured list of select tags
+      .then((response) => {
+        return { data: excludeMetaTag('tagset', response.data) };
+      });
```
jujaga, line 32: Can use closures here:
```ts
.then((response) => ({ data: excludeMetaTag('tagset', response.data) }));
```

frontend/src/services/versionService.ts
```diff
@@ -11,7 +12,11 @@ export default {
    * @returns {Promise} An axios response
    */
   getMetadata(headers: any = {}, params: GetVersionMetadataOptions) {
-    return comsAxios().get(`${PATH}/metadata`, { headers: headers, params: params });
+    return comsAxios().get(`${PATH}/metadata`, { headers: headers, params: params })
+      // filter out a configured list of select metadata
+      .then((response) => {
+        return { data: excludeMetaTag('metadata', response.data) };
+      });
```
jujaga, line 19: Can use closures here:
```ts
.then((response) => ({ data: excludeMetaTag('metadata', response.data) }));
```

frontend/src/services/objectService.ts
```diff
@@ -287,7 +299,11 @@ export default {
       params: {
         tagset: Object.fromEntries((tagset.map((x: { key: string; value: string }) => ([x.key, x.value]))))
       }
-    });
+    })
+      // filter out a configured list of select tags
+      .then((response) => {
+        return { data: excludeMetaTag('tagset', response.data) };
+      });
```
jujaga, line 306: Can use closures here:
```ts
.then((response) => ({ data: excludeMetaTag('tagset', response.data) }));
```

frontend/src/services/objectService.ts
```diff
@@ -215,7 +223,11 @@ export default {
         ...Object.fromEntries((headers.metadata.map((x: { key: string; value: string }) => ([x.key, x.value]))))
       };
     }
-    return comsAxios().get(`${PATH}/metadata`, config);
+    return comsAxios().get(`${PATH}/metadata`, config)
+      // filter out a configured list of select metadata
+      .then((response) => {
+        return { data: excludeMetaTag('metadata', response.data) };
+      });
```
jujaga, line 230: Can use closures here:
```ts
.then((response) => ({ data: excludeMetaTag('metadata', response.data) }));
```

frontend/src/services/objectService.ts
```diff
@@ -112,7 +116,11 @@ export default {
    * @returns {Promise} An axios response
    */
   getObjectTagging(params: GetObjectTaggingOptions = {}) {
-    return comsAxios().get(`${PATH}/tagging`, { params: params });
+    return comsAxios().get(`${PATH}/tagging`, { params: params })
+      // filter out a configured list of select tags
+      .then((response) => {
+        return { data: excludeMetaTag('tagset', response.data) };
+      });
```
jujaga, line 123: Can use closures here:
```ts
.then((response) => ({ data: excludeMetaTag('tagset', response.data) }));
```

frontend/src/services/objectService.ts
```diff
@@ -102,7 +102,11 @@ export default {
   getMetadata(headers: any = {}, params: GetMetadataOptions = {}) {
     // remove objectId array if its first element is undefined
     if (params.objectId && params.objectId[0] === undefined) delete params.objectId;
-    return comsAxios().get(`${PATH}/metadata`, { headers: headers, params: params });
+    return comsAxios().get(`${PATH}/metadata`, { headers: headers, params: params })
+      // filter out metadata and return as a promise again
+      .then((response) => {
+        return { data: excludeMetaTag('metadata', response.data) };
+      });
```
jujaga, line 109: Can use closures here:
```ts
.then((response) => ({ data: excludeMetaTag('metadata', response.data) }));
```

frontend/src/components/object/ObjectMetadata.vue
```diff
@@ -97,7 +97,10 @@ watch([props, tsGetMetadata,vsGetMetadata] , () => {
 </script>
 
 <template>
-  <div class="grid details-grid grid-nogutter mb-2">
+  <div
+    v-if="objectMetadata?.metadata.length"
```
kyle1morel, line 101: Double check with Tyler. I believe I didn't v-if this intentionally. Something about wanting to show the object had no metadata.

kyle1morel (approved):

jujaga (approved):

jatindersingh93 (approved):

PR #139, opened 2023-10-20
adjust and remove heading css

jatindersingh93:
Removed, adjust and reuse heading tags.
This should fix reported formatting issue to login/logout page and other unreported inconsistency.

# Description

We have to make sure all headings are being displayed properly in terms of size and location on screen.
This should fix login/logout and other inconsistency caused by adjusting heading tags.
## Checklist

### DIFF
--- a/frontend/src/assets/primevue.scss
+++ b/frontend/src/assets/primevue.scss
@@ -118,3 +118,8 @@
   padding: .25rem .75rem;
   font-weight: normal;
 }
+
+h1, h2, h3, h4, h5 {
+	font-weight: 600;
+	line-height: 1.2em;
+}
--- a/frontend/src/components/bucket/BucketList.vue
+++ b/frontend/src/components/bucket/BucketList.vue
@@ -132,9 +132,6 @@ onMounted(async () => {
 </template>
 
 <style lang="scss" scoped>
-h1 {
-  font-weight: bold;
-}
 
 button {
   text-indent: 10px;
--- a/frontend/src/components/bucket/BucketSidebar.vue
+++ b/frontend/src/components/bucket/BucketSidebar.vue
@@ -87,7 +87,7 @@ watch(props, () => {
   <div class="pl-2 sidebar">
     <div class="grid details-grid grid-nogutter">
       <div class="col-12">
-        <h2 class="font-bold">
+        <h2>
           Properties
         </h2>
       </div>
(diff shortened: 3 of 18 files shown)


TimCsaky:
the changes look good. those pages were pretty hidden!

more of a visual/css issue for me but, i wonder if these h1 headings need to be bold, others h1's are font-weight 400. 
or maybe they all should be.
maybe the one in the sidebar should be smaller.
image

this heading as well is maybe inconsistent.
image
maybe we ask @tpantella

tpantella:
> maybe we ask @tpantella

Hi! I think "Select a bucket" looks correct, but whatever more closely matches h1 on gov.bc.ca, such as this example:
https://www2.gov.bc.ca/gov/content/health/health-drug-coverage/pharmacare-for-bc-residents/drug-review-process-results/drug-review-decisions
"Drug review decisions" is the correct styling and weight. We should try to mimic this with our H1s.  You can jump around on gov.bc.ca to see examples of H2 and H3 as well if that is of interest.

On an object list, the "Files" heading seems to be the wrong weight, it should be consistent with the above.
On a full file details page, the "File details" heading seems to be the wrong weight, it should match the above.

On the side pane heading, which is what Tim was actually asking about, I believe it's ok to have H1s in the side pane - it's ok to use multiple H1s, the side pane is acting as a sort of framed page, and it's a separate and overarching topic that earns an H1. I could also see a case for it being H2, which would then force any child headings to be H3, and children of children to be H4. My preference is H1 since it must draw attention to itself and becomes the new "main topic" being observed on the page. If a developer has a strong preference and makes a case for it, or if H2 for side pane headers and cascading all other sizes down one is easier and looks good, I am open to that. This comes down more to what looks good, since there is a style justification for either approach.

Design System guidance on weights is here if this helps:
https://developer.gov.bc.ca/Design-System/Typography

I think specified weights rather than specified bold styling is interpreted differently by screen reader software, where bold can invoke a shouty voice or emphasis, and I think we should also try to avoid variants of headings in bold and non-bold at the same H-level. 

tl;dr - I think we need to make all H1s heavy in weight but not explicitly stated as "bold", I think some need fixing, I think the side pane heading could be H1 or H2 depending on what looks best, some adjustments may be needed on children headings if it goes to H2

TimCsaky (approved):

jujaga (approved):

PR #142, opened 2023-10-27
Sc3171

jatindersingh93:
# Description
- Provides clear button (X) inside the search input box
- Include Tooltips to every icon button including clear button
- Corrections to Aria-label specifically to syn and detail buttons
- Make sure search clear button can be navigated using tab and consistent with focus change

## Checklist

### DIFF
--- a/frontend/src/assets/main.scss
+++ b/frontend/src/assets/main.scss
@@ -183,5 +183,4 @@ td .wrap-block {
 .p-input-icon-clear-right {
   position: absolute !important;
   right: 0rem;
-  line-height: 1.2;
 }
--- a/frontend/src/assets/main.scss
+++ b/frontend/src/assets/main.scss
@@ -180,3 +180,9 @@ td .wrap-block {
     cursor: pointer;
   }
 }
+.p-input-icon-clear-right {
+  position: absolute !important;
+  right: 0px;
+  line-height: 1.2;
+
+}
--- a/frontend/src/assets/main.scss
+++ b/frontend/src/assets/main.scss
@@ -182,7 +182,6 @@ td .wrap-block {
 }
 .p-input-icon-clear-right {
   position: absolute !important;
-  right: 0px;
+  right: 0rem;
   line-height: 1.2;
-
 }
--- a/frontend/src/components/object/ObjectTable.vue
+++ b/frontend/src/components/object/ObjectTable.vue
@@ -96,8 +96,6 @@ const filters = ref({
   // @ts-ignore
   global: { value: null, matchMode: FilterMatchMode.CONTAINS }
 });
-
-console.log(filters)
 </script>
 
 <template>
@@ -131,6 +129,7 @@ console.log(filters)
             />
             <Button
               v-show="filters['global'].value != null"
+              v-tooltip.bottom="'Clear'"
               class="ml-2 p-input-icon-clear-right"
               icon="pi pi-times"
               outlined
@@ -138,7 +137,9 @@ console.log(filters)
               @click="filters['global'].value = null"
             />
           </span>
+
           <Button
+            v-tooltip.bottom="'Refresh'"
             class="ml-2"
             icon="pi pi-refresh"
             outlined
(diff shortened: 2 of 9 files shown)


frontend/src/assets/main.scss
```diff
@@ -180,3 +180,8 @@ td .wrap-block {
     cursor: pointer;
   }
 }
+.p-input-icon-clear-right {
+  position: absolute !important;
+  right: 0rem;
+  line-height: 1.2;
```
kyle1morel, line 186: Probably want to make this fill the whole space instead of having that tiny gap below it. Removing line-height should do this. Could also do height: 100% as a safety net.

frontend/src/components/object/ObjectTable.vue
```diff
@@ -225,20 +235,25 @@ const filters = ref({
...
+          />
           <Button
             v-if="
               data.public ||
               permissionStore.isObjectActionAllowed(data.id, getUserId, Permissions.READ, props.bucketId as string)
             "
+            v-tooltip.bottom="'Bucket detail'"
             class="p-button-lg p-button-rounded p-button-text"
-            aria-label="Synchronize bucket"
+            aria-label="Bucket detail"
```
kyle1morel, line 256: Should be 'Object details'

frontend/src/components/object/ObjectTable.vue
```diff
@@ -225,20 +235,25 @@ const filters = ref({
...
+          <SyncButton
+            label-text="Synchronize file"
+            :object-id="data.id"
+          />
           <Button
             v-if="
               data.public ||
               permissionStore.isObjectActionAllowed(data.id, getUserId, Permissions.READ, props.bucketId as string)
             "
+            v-tooltip.bottom="'Bucket detail'"
```
TimCsaky, line 254: this should be 'Object details' as well

frontend/src/components/object/ObjectTable.vue
```diff
@@ -127,9 +127,19 @@ const filters = ref({
               v-model="filters['global'].value"
```
TimCsaky, line 127: minor point but to avoid over-powering css, it might be worth adding a class to this element so the css rule below is more targeted:
`<InputText class="searchInput"..`
...
then on line 297:
```css
:deep(.searchInput .p-inputtext) { /* might be .searchInput:deep(p.inputtext) { */
  padding-right: 2.5rem;
}
```

TimCsaky (approved):
looks good to me.

frontend/src/assets/main.scss
jatindersingh93: `line-height` is removed. it seems to be inheriting height.

frontend/src/components/object/ObjectTable.vue
```diff
@@ -251,9 +252,9 @@ const filters = ref({
               data.public ||
               permissionStore.isObjectActionAllowed(data.id, getUserId, Permissions.READ, props.bucketId as string)
             "
-            v-tooltip.bottom="'Bucket detail'"
+            v-tooltip.bottom="'Bucket details'"
             class="p-button-lg p-button-rounded p-button-text"
-            aria-label="Bucket detail"
+            aria-label="Bucket details"
```
kyle1morel, line 257: Still says Bucket instead of Object

jatindersingh93: Ya, I see some inconsistency as well. Do you think it should be "file details" instead. This will make it consistent with the popup?

kyle1morel (approved):

frontend/src/components/object/ObjectTable.vue
```diff
@@ -125,11 +125,22 @@ const filters = ref({
             <i class="pi pi-search" />
             <InputText
               v-model="filters['global'].value"
+              class="searchInput"
               placeholder="Search File Names"
             />
+            <Button
+              v-show="filters['global'].value != null"
```
jujaga, line 132: !== stronger equality

PR #146, opened 2023-11-14
Implement support for nested buckets view

kyle1morel:
# Description

Changes the bucket list view to use a TreeTable instead of a DataTable. This allows us to mimic a folder hierarchy of buckets based on their canonical path (endpoint, bucket, key). Fake folders will be inserted to create a full tree if there is no direct parent bucket available. 

This PR also slipstream resolves a build issue with the Dockerfile.

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/types/BucketTreeNode.ts
+++ b/frontend/src/types/BucketTreeNode.ts
@@ -1,0 +1,16 @@
+import type { Bucket } from '@/types';
+
+export type BucketTreeNode = {
+  key: string;
+  data: Bucket | BucketTreeNodeDummyData;
+  isRoot: boolean;
+  children: Array<BucketTreeNode>;
+};
+
+export type BucketTreeNodeDummyData = {
+  bucket: string;
+  bucketName: string;
+  dummy: boolean;
+  endpoint: string;
+  key: string;
+};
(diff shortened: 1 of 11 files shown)


frontend/src/types/BucketTreeNode.ts
```diff
@@ -0,0 +1,16 @@
+import type { Bucket } from '@/types';
+
+export type BucketTreeNode = {
+  key: string;
+  data: Bucket | BucketTreeNodeDummyData;
```
jujaga, line 5: It may generally be better to coerce this type to be one explicit type assignmenet instead of a union and defer the intersectioning interface to the child type instead - that way you should be able to theoretically avoid needing to do any casting in the code where we actually leverage this. See comment below.
```ts
data: BucketTreeNodeData
```

frontend/src/utils/utils.ts
```diff
@@ -13,6 +13,25 @@ export function differential(source: any, comparer: any): any {
...
 }
 
+/**
+ * @function isAtPath
+ * Predicate function determining if the `path` is a non-directory member of the `prefix` path
+ * @param {string} prefix The base "folder"
+ * @param {string} path The "file" to check
+ * @returns {boolean} True if path is member of prefix. False in all other cases.
+ */
+export function isAtPath(prefix: string, path: string) {
```
jujaga, line 23: Return type of this function isn't typescript defined
```ts
export function isAtPath(prefix: string, path: string) boolean {
```

jujaga (approved):

TimCsaky (approved):

norrisng-bc (approved):

PR #148, opened 2023-11-21
CSS refactor to be more compatible with PrimeVue and improve layout on smaller screens 

TimCsaky:
support ticket:
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3384

consolidate our style sheets
resolve style conflicts
use primevue defaults where possible
make semi-responsive

# Description

## Checklist


tpantella:
New styling is imposing a bunch of white space on the right that looks off. Suggest matching the current Prod BCBox styling more closely when it comes to large browser resolutions. I'm at 2560 x 1440 using Chrome at 100% zoom. Although I did always think Prod BCBox was a bit -wide- so the main issue is that it is not centered - it would be more appropriate if the white space was distributed between left and right. So my advice is actually to slightly widen the content / reduce the margins, but also center the content.

Line spacing appears to be more compact compared to the current Prod BCBox, suggest more space between rows of text to more closely match current Prod BCBox.

image

I see the same margin/centering issue on the bucket list, but even more pronounced - the datatable is very narrow and this should be fixed.

image

This margin/centering issue is also true for the object list. I have also noted that Upload/Download don't seem to have hover tooltips, but Delete does. I like the border width for buttons.

Otherwise looks good to me!

TimCsaky:
> New styling is imposing a bunch of white space on the right that looks off. Suggest matching the current Prod BCBox styling more closely when it comes to large browser resolutions. I'm at 2560 x 1440 using Chrome at 100% zoom. Although I did always think Prod BCBox was a bit -wide- so the main issue is that it is not centered - it would be more appropriate if the white space was distributed between left and right. So my advice is actually to slightly widen the content / reduce the margins, but also center the content.
> 
> Line spacing appears to be more compact compared to the current Prod BCBox, suggest more space between rows of text to more closely match current Prod BCBox.
> 
> image
> 
> I see the same margin/centering issue on the bucket list, but even more pronounced - the datatable is very narrow and this should be fixed.
> 
> image
> 
> This margin/centering issue is also true for the object list. I have also noted that Upload/Download don't seem to have hover tooltips, but Delete does. I like the border width for buttons.
> 
> Otherwise looks good to me!

Thanks Tyler. this should be updated now: https://bcbox-dev-pr-148.apps.silver.devops.gov.bc.ca/

jujaga (approved):

frontend/src/components/object/ObjectTable.vue
```diff
@@ -157,7 +157,7 @@ const filters = ref({
           v-if="!useAppStore().getIsLoading"
           class="flex justify-content-center"
         >
-          <h3>There are no objects associated with your account in this bucket.</h3>
+          <h4>There are no objects associated with your account in this bucket.</h4>
```
jatindersingh93, line 160: Isn't this should be h3? the order should be h1, h2 and h3

frontend/src/components/object/ObjectVersion.vue
```diff
@@ -183,8 +184,12 @@ watch(getVersions, async () => {
                 )
               "
               :to="{ name: RouteNames.DETAIL_OBJECTS, query: { objectId: props.objectId, versionId: data.id } }"
+              class="action-link"
             >
-              <Button class="p-button-lg p-button-rounded p-button-text">
+              <Button
+                class="p-button-lg p-button-rounded p-button-text"
+                v-tooltip.bottom="'Version details'"
```
jatindersingh93, line 191: some unit test errors, that needs to be fixed.

kyle1morel:
The very first image on the main page, under the _Login to get started_ button has a bunch of padding in the image not present in live, or the other images on the pr.

frontend/tests/unit/components/layout/Footer.spec.ts
```diff
@@ -40,14 +40,15 @@ describe('Footer.vue', () => {
...
+  it('contains 7 links', () => {
     const wrapper = mount(Footer, {
       global: {
         plugins: [createTestingPinia(), PrimeVue]
       }
     });
 
-    const btn = wrapper.findAll('button');
-    expect(btn).toHaveLength(7);
+    console.log('a', wrapper);
```
jatindersingh93, line 50: another unit test errors

jatindersingh93 (approved):

norrisng-bc (approved):
Not sure if intentional, but the checkboxes (e.g. in the permissions modal) seem to be using the Primevue default light blue instead of the BC Gov dark blue (`$bcbox-primary: #036` inside `variables.scss`).

PR #149, opened 2023-12-01
ObjectTable: select all on current page only

### LINKED ISSUE #112: Select all toggle, selects ALL items in BC Box (including not visible items on other pages)
#### Describe the bug
When you toggle the "select all" radio button, it selects all the items on the BC Box bucket. Applying any action to this selection applies it to everything in the bucket.

ie. Deleting (a page of items) deletes all the items in the bucket

#### Expected behavior
Apply action to the selected items on the current page (not the entire bucket)
This would match the UX of other sites, as to not delete items that are not visible on the current page

norrisng-bc:
# Description

This PR resolves #112.

The "select all" checkbox in the objects table now only selects items visible on the current page.

Primevue doesn't provide a native way to do this at the moment (see primefaces/primevue/issues/3477), so the solution involved figuring out what objects are on the current page.

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3365

## Types of changes

## Checklist

## Further comments

The DOM had to be directly queried as a workaround.

For the most part, the list of objects provided by `DataTable`'s `value-change` emit is correctly sorted, in that it matches what's currently displayed in the table.

However, this is _not_ the case on initial page load. Instead, `value-change` simply uses the order as returned by the COMS API (specifically, by `GET /object`); however, the `DataTable` default sort order (and therefore the display order on initial page load) is the last updated date, which differs from the order returned by the COMS API.

Since the updated "select all" behaviour works by selecting the top row plus the next `{number of items per page}` objects (both of these are available via the `on-page` emit), a discrepancy in object order will result in the "wrong" items being selected (i.e. some objects on the current page don't get selected, while others on other pages do).

As a workaround, a DOM query on the `DataTable` is used to determine the actual display order. This is made possible by adding a custom HTML attribute (`data-objectid`) containing the object ID in each table row.

### DIFF
--- a/frontend/src/components/object/ObjectTable.vue
+++ b/frontend/src/components/object/ObjectTable.vue
@@ -46,6 +46,7 @@ const { getObjects } = storeToRefs(objectStore);
 const { getUserId } = storeToRefs(useAuthStore());
 
 // State
+const selectAll = ref(false);
 const permissionsVisible = ref(false);
 const permissionsObjectId = ref('');
 const permissionsObjectName: Ref<string | undefined> = ref('');
@@ -54,27 +55,37 @@ const tableData: Ref<Array<COMSObjectDataSource>> = ref([]);
 // Actions
 const toast = useToast();
 
-const formatShortUuid = (uuid: string) => {
-  return uuid?.slice(0, 8) ?? uuid;
-};
+function clearSelected() {
+  selectAll.value = false;
+  objectStore.setSelectedObjects([]);
+}
 
-const showInfo = async (id: string) => {
-  emit('show-object-info', id);
-};
+const formatShortUuid = (uuid: string) => uuid?.slice(0, 8) ?? uuid;
+const onDeletedSuccess = () => toast.success('File deleted');
 
-const showPermissions = async (objectId: string) => {
+function selectCurrentPage() {
+  objectStore.setSelectedObjects(
+    tableData.value.filter((object) => {
+      return Array.from(document.querySelectorAll('[data-objectId]'))
+        .map((x) => x.getAttribute('data-objectId'))
+        .includes(object.id);
+    })
+  );
+}
+
+const showInfo = (id: string) => emit('show-object-info', id);
+
+async function showPermissions(objectId: string) {
   await permissionStore.fetchObjectPermissions({ objectId });
 
   permissionsVisible.value = true;
   permissionsObjectId.value = objectId;
   permissionsObjectName.value = objectStore.findObjectById(objectId)?.name;
-};
-
-function onDeletedSuccess() {
-  toast.success('File deleted');
 }
 
 watch(getObjects, async () => {
+  clearSelected();
+
   // Filter object cache to this specific bucket
   const objs: Array<COMSObjectDataSource> = getObjects.value.filter(
     (x: COMSObject) => x.bucketId === props.bucketId
@@ -118,6 +129,21 @@ const filters = ref({
       sort-field="lastUpdatedDate"
       :sort-order="-1"
       :global-filter-fields="['name']"
+      :select-all="selectAll"
+      @select-all-change="
+        (e) => {
+          selectAll = e.checked;
+          if (e.checked) {
+            selectCurrentPage();
+          } else {
+            objectStore.setSelectedObjects([]);
+          }
+        }
+      "
+      @page="clearSelected"
+      @update:sort-order="clearSelected"
+      @update:sort-field="clearSelected"
+      @update:rows="clearSelected"
     >
       <template #header>
         <div class="flex justify-content-end">
@@ -129,6 +155,7 @@ const filters = ref({
               v-model="filters['global'].value"
               class="searchInput"
               placeholder="Search File Names"
+              @input="clearSelected()"
             />
             <Button
               v-show="filters['global'].value !== null"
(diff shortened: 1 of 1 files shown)


frontend/src/components/object/ObjectTable.vue
```diff
@@ -54,6 +54,39 @@ const tableData: Ref<Array<COMSObjectDataSource>> = ref([]);
...
+ * @function unselectAllObjects
+ * Helper function for unselecting all objects
+ */
+function unselectAllObjects() {
+  objectStore.setSelectedObjects([]);
+  selectAll.value = false;
+}
+
+/**
+ * @function queryTableOrder
```
TimCsaky, line 80: not sure about thew name of this function. maybe `selectAllVisible()`?

frontend/src/components/object/ObjectTable.vue
```diff
@@ -175,7 +257,10 @@ const filters = ref({
         body-class="truncate"
       >
         <template #body="{ data }">
-          <div v-tooltip.bottom="{ value: data.name }">
+          <div
+            v-tooltip.bottom="{ value: data.name }"
+            :data-objectid="data.id"
```
TimCsaky, line 262: feels a bit precarious adding this attribute to the tooltip. maybe it should appear in a hidden column.
i wonder if it's also available from the dataKey for each row (https://primevue.org/datatable/#api.datatable.props.dataKey)

TimCsaky (commented):
If we're going to get visible rows by inspecting the DOM maybe you dont need to look at first and page. just do 'data.filter(obj => domIds.includes(obj.id)'
just trying to make things easier.. not sure if it's as simple as that

frontend/src/components/object/ObjectTable.vue
```diff
@@ -54,6 +54,39 @@ const tableData: Ref<Array<COMSObjectDataSource>> = ref([]);
...
+
+// list of objects, in the order dislpayed in the DataTable
+const objectsList: Ref<Array<COMSObject>> = ref([]);
+
+// Initial page load?
+const isInitialLoad = ref(true);
+
+// DataTable pagination state
+const itemsPerPage = ref(10);
+const currentPageTop = ref(0);
```
jujaga, line 68: Do we have these magic numbers as constants somewhere? I know we generally use 10 as our default pagination size, but that should be defined as a constant across the codebase so that we're not just raw-defining default integer states.

frontend/src/components/object/ObjectTable.vue
```diff
@@ -111,13 +188,18 @@ const filters = ref({
...
       current-page-report-template="{first}-{last} of {totalRecords}"
       :rows-per-page-options="[10, 20, 50]"
       sort-field="lastUpdatedDate"
       :sort-order="-1"
       :global-filter-fields="['name']"
+      :select-all="selectAll"
+      @select-all-change="onSelectAllChange"
+      @page="onPage"
+      @value-change="onValueChange"
+      @sort="onSort"
```
jujaga, line 202: Style: Let's make sure the attributes are alphabetically sorted if possible.

frontend/src/components/object/ObjectTable.vue
```diff
@@ -98,6 +131,50 @@ const filters = ref({
...
+  }
+};
+
+const onPage = (event: any) => {
+  currentPageTop.value = event.first;
+  itemsPerPage.value = event.rows;
+  unselectAllObjects();
+};
+
+const onValueChange = (event: any) => {
```
jujaga, line 171: Is there any way we can be more specific about the type of event that will be coming in? Sometimes the `any` keyword is necessary, but that is usually indicative of either a breakdown of typescript, or the actual data structure expectations haven't been thoroughly investigated yet.

frontend/src/components/object/ObjectTable.vue
```diff
@@ -98,6 +131,50 @@ const filters = ref({
...
 });
+
+const onSelectAllChange = (event: any) => {
+  selectAll.value = event.checked;
+  if (event.checked) {
+    if (isInitialLoad.value) {
+      /*
+       * On initial page load, the object order returned by this emit doesn't corespond
+       * to the actual display order - it's just the order returned by the COMS API.
+       * Workaround: directly query the DOM to figure out the actual display order (see queryTableOrder()).
```
jujaga, line 142: Isn't it the datatable component's job to be managing the order? I'm not completely sure why we need to be doing this, unless we need the explicit data ordering value in this parent component and there isn't any other way to pull that information from the datatable API?

frontend/src/components/object/ObjectTable.vue
norrisng-bc: It was previously a magic number as well, as `<DataTable :rows:"10">`. There's `:rows-per-page-options="[10, 20, 50]"` but otherwise that's it afaik.

frontend/src/components/object/ObjectTable.vue
norrisng-bc: Initial page load is an edge case - typically, the event returned by the `value-change` emit is in sorted order, but only on page load does it return the order from the COMS API instead of what's displayed. If you paginate, change sort order etc after this initial load it'll be in the correct order though.

kyle1morel (approved):

jujaga (approved):

PR #151, opened 2023-12-08
Fix adding multiple users to buckets and objects

wilwong89:
# Description
Giving permissions to one user will no longer remove other users without permissions from the permissions list. For objects and buckets

When user is added to permissions, they automatically have read perms 

Modified files:
	BucketPermission.vue
	BucketPermissionAddUser.vue
	ObjectPermission.vue
	ObjectPermissionAddUser.vue
	permissionStore.ts
## Types of changes

## Checklist

### DIFF
--- a/frontend/src/store/permissionStore.ts
+++ b/frontend/src/store/permissionStore.ts
@@ -66,14 +66,6 @@ export const usePermissionStore = defineStore('permission', () => {
     }
   }
 
-  function addBucketUser(user: UserPermissions) {
-    try {
-      getters.getMappedBucketToUserPermissions.value.push(user);
-    } catch (error: any) {
-      toast.error('Adding bucket user', error);
-    }
-  }
-
   async function addObjectPermission(objectId: string, userId: string, permCode: string) {
     try {
       appStore.beginIndeterminateLoading();
@@ -89,14 +81,6 @@ export const usePermissionStore = defineStore('permission', () => {
       await fetchObjectPermissions();
       await mapObjectToUserPermissions(objectId);
       appStore.endIndeterminateLoading();
-    }
-  }
-
-  function addObjectUser(user: UserPermissions) {
-    try {
-      getters.getMappedObjectToUserPermissions.value.push(user);
-    } catch (error: any) {
-      toast.error('Adding object user', error);
     }
   }
 
@@ -327,9 +311,7 @@ export const usePermissionStore = defineStore('permission', () => {
 
     // Actions
     addBucketPermission,
-    addBucketUser,
     addObjectPermission,
-    addObjectUser,
     deleteBucketPermission,
     deleteObjectPermission,
     fetchBucketPermissions,
--- a/frontend/src/store/permissionStore.ts
+++ b/frontend/src/store/permissionStore.ts
@@ -53,7 +53,7 @@ export const usePermissionStore = defineStore('permission', () => {
   };
 
   // Actions
-  async function addBucketPermission(bucketId: string, userId: string, permCode: string) {
+  async function addBucketPermission(bucketId: string, userId: string, permCode: string): Promise<void> {
     try {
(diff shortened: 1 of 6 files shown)


frontend/src/store/permissionStore.ts
```diff
@@ -186,13 +202,13 @@ export const usePermissionStore = defineStore('permission', () => {
     }
   }
 
-  function isBucketActionAllowed(bucketId: string, userId?: string, permCode?: string) {
+  function isBucketActionAllowed(bucketId: string, userId?: string, permCode?: string): Boolean {
```
jujaga, line 205: Type should be lowercase `boolean` primitive, not the boxed variant `Boolean`

frontend/src/store/permissionStore.ts
```diff
@@ -186,13 +202,13 @@ export const usePermissionStore = defineStore('permission', () => {
...
     return state.bucketPermissions.value.some(
       (x: BucketPermission) => x.bucketId === bucketId && x.userId === userId && x.permCode === permCode
     );
   }
 
-  function isObjectActionAllowed(objectId: string, userId?: string, permCode?: string, bucketId?: string) {
+  function isObjectActionAllowed(objectId: string, userId?: string, permCode?: string, bucketId?: string): Boolean {
```
jujaga, line 211: Type should be lowercase `boolean` primitive, not the boxed variant `Boolean`

frontend/src/store/permissionStore.ts
```diff
@@ -204,15 +220,15 @@ export const usePermissionStore = defineStore('permission', () => {
     return bucketPerm || objectPerm;
   }
 
-  function isUserElevatedRights() {
+  function isUserElevatedRights(): Boolean {
```
jujaga, line 223: Type should be lowercase `boolean` primitive, not the boxed variant `Boolean`

frontend/src/store/permissionStore.ts
```diff
@@ -87,33 +94,39 @@ export const usePermissionStore = defineStore('permission', () => {
       toast.error('Adding object permission', error);
     } finally {
       await fetchObjectPermissions();
+
+      const initialPerms = [...state.mappedObjectToUserPermissions.value];
       await mapObjectToUserPermissions(objectId);
+      const afterPerms = [...state.mappedObjectToUserPermissions.value];
+
+      const results = initialPerms.filter(({ userId: id1 }) => !afterPerms.some(({ userId: id2 }) => id2 === id1));
+      state.mappedObjectToUserPermissions.value = state.mappedObjectToUserPermissions.value.concat(results);
```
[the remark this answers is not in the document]
wilwong89: The data returned by line 91 does not return the full details, it returns a different object that is slightly different. For example, the data returned does not have the name or email fields and if the previous user permission did not already have the READ perm code and another (non-READ) perm code is added, the data returned suggests that only READ was updated.

frontend/src/store/permissionStore.ts
```diff
@@ -293,13 +309,23 @@ export const usePermissionStore = defineStore('permission', () => {
     } catch (error: any) {
       toast.error('Removing bucket user', error);
     } finally {
-      await fetchBucketPermissions();
-      await mapBucketToUserPermissions(bucketId);
+      await removeBucketUserPermsHandler(bucketId, userId);
```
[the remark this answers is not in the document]
wilwong89: Using your method to read the deleted permission and then dropping it from the local state requires storing an array of deleted permissions returned form line, iteratively checking that list and removing to remove the permissions from the local state, and then finally another check to see if all permissions were removed from an entry in the local list before removing the entry entirely. I think instead the current method of doing a simple diff from before and after is the better method.

kyle1morel (approved):

jujaga (approved):

jatindersingh93 (approved):

PR #152, opened 2023-12-08
Create toast if user updates credentials for bucket that has children

norrisng-bc:
# Description
This PR adds an additional toast, whenever credentials are updated for a bucket that has children:

Screenshot 2023-12-08 at 12 39 29 PM

The additional toast is displayed for 10 seconds.

Although the toast asks the user if there's any child buckets, it's only displayed if there *are* child buckets managed inside COMS/BCBox.

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3372

## Types of changes

## Checklist

## Further comments

The single quotes in the toast message are conflicting with the linter:
* If escaping the single quotes within a single quote delineated string (i.e `'Don\'t forget...'`), the linter will convert it back into double quotes and remove the escape sequence (i.e. `"Don't forget..."`
* If using double quotes without escaping `\'`, it results in a linter error

### DIFF
--- a/frontend/src/components/bucket/BucketConfigForm.vue
+++ b/frontend/src/components/bucket/BucketConfigForm.vue
@@ -100,11 +100,7 @@ const onSubmit = async (values: any) => {
     toast.success('Configuring bucket', 'Bucket configuration successful');
 
     if ((bucketChanges.accessKeyId || bucketChanges.secretAccessKey) && hasChildren) {
-      toast.info(
-        'Any child buckets?',
-        "Don't forget to update their credentials too, if you're managing them in BCBox",
-        { life: 10000 }
-      );
+      toast.info('Buckets under this bucket', 'Remember to update their credentials where applicable', { life: 10000 });
     }
   } catch (error: any) {
     toast.error('Configuring bucket', error);
--- a/frontend/src/components/bucket/BucketConfigForm.vue
+++ b/frontend/src/components/bucket/BucketConfigForm.vue
@@ -7,7 +7,7 @@ import Password from '@/components/form/Password.vue';
 import TextInput from '@/components/form/TextInput.vue';
 import { Button, useToast } from '@/lib/primevue';
 import { useAuthStore, useBucketStore } from '@/store';
-import { differential, joinPath } from '@/utils/utils';
+import { differential, getBucketPath, joinPath } from '@/utils/utils';
 
 import type { Bucket } from '@/types';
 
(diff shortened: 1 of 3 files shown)


frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -77,14 +77,35 @@ const onSubmit = async (values: any) => {
...
-      ? await bucketStore.updateBucket(props.bucket?.bucketId, differential(formBucket, initialValues))
+      ? await bucketStore.updateBucket(props.bucket?.bucketId, bucketChanges)
       : await bucketStore.createBucket(formBucket);
 
     await bucketStore.fetchBuckets({ userId: getUserId.value, objectPerms: true });
+
+    // trim trailing "//", if present
+    // (modifying getBucketPath() instead seems to break nested bucket display [PR-146])
+    const currBucketPath = getBucketPath(initialValues as Bucket).endsWith('//')
+      ? getBucketPath(initialValues as Bucket).slice(0, -1)
```
TimCsaky, line 91: dont you need to call `getBucketPath(formBucket as Bucket)` using the `formBucket` (updated properties) rather than initialValues?

norrisng-bc: `getBucketPath()` expects the bucket key, which `formBucket` doesn't have since you aren't really (nor can you) change the path for an existing COMS bucket. If the bucket doesn't point at a subdirectory in the bucket you'll end up with `undefined` tacked onto the end where it expects the bucket sub-path (e.g. `https://s3.amazonaws.com/yxwgj/undefined`).

That helper function was lifted as-is from `BucketTable` into `utils.ts`, so I'm a bit wary of changing how it works to accommodate this.

TimCsaky: sorry. i forgot you can't edit the bucket key. initialValues is fine

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -77,14 +77,31 @@ const onSubmit = async (values: any) => {
...
+
     props.bucket
-      ? await bucketStore.updateBucket(props.bucket?.bucketId, differential(formBucket, initialValues))
+      ? await bucketStore.updateBucket(props.bucket?.bucketId, bucketChanges)
       : await bucketStore.createBucket(formBucket);
 
     await bucketStore.fetchBuckets({ userId: getUserId.value, objectPerms: true });
+
+    // trim trailing "//", if present
+    // (modifying getBucketPath() instead seems to break nested bucket display [PR-146])
```
TimCsaky, line 89: i see the issue here. we could modify the getBucketPath function to remove the trailing slash.. but that will impact the other uses of that function.

TimCsaky (approved):
looks good to me.

frontend/src/components/bucket/BucketConfigForm.vue
```diff
@@ -77,14 +77,31 @@ const onSubmit = async (values: any) => {
...
+    const hasChildren = bucketStore.buckets.some(
+      (b) => getBucketPath(b).includes(currBucketPath) && getBucketPath(b) !== currBucketPath
+    );
+
     emit('submit-bucket-config');
 
     toast.success('Configuring bucket', 'Bucket configuration successful');
+
+    if ((bucketChanges.accessKeyId || bucketChanges.secretAccessKey) && hasChildren) {
+      toast.info('Buckets under this bucket', 'Remember to update their credentials where applicable', { life: 10000 });
```
TimCsaky, line 103: assuming this wording was approved by our ux folks.
it's certainly a hard one to word

jujaga (approved):

PR #164, opened 2024-01-19
Refactor DataTable with Pagination SearchObject API service call

jatindersingh93:
- Cleanly structured code (eg. dataTable component, objectStore, objectService) that is documented
- A pattern defined
- object table fetches only 1 page (eg 10, 20 or 50) of records using a direct call to the COMS searchObjects endpoint
- the object data in the response from COMS could be added to the store but the process of fetching does not read from the store
- permissions data should remain available (from the store or wherever it is pulled by other components). The action buttons in the object table should still be reading from permissions in the store.
- clicking through page numbers below the datatable will calls COMS again to fetch the next set of records.
- the total number of pages can be displayed using the response X-Total-rows header
- Modify the primevue DataTable implemented in the objectTable component to get data directly from COMS api searchObjects endpoint instead of reading object data in the store

[SC-3467](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3467)

# Description

## Checklist

### DIFF
--- a/frontend/src/components/object/ObjectTable.vue
+++ b/frontend/src/components/object/ObjectTable.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { storeToRefs } from 'pinia';
-import { onUnmounted, ref, watch } from 'vue';
+import { onUnmounted, onMounted, ref } from 'vue';
 
 import { Spinner } from '@/components/layout';
 import {
@@ -13,10 +13,11 @@ import {
 import { SyncButton } from '@/components/common';
 import { ShareObjectButton } from '@/components/object/share';
 import { Button, Column, DataTable, Dialog, FilterMatchMode, InputText, useToast } from '@/lib/primevue';
-import { useAuthStore, useAppStore, useObjectStore, usePermissionStore } from '@/store';
+import { useAuthStore, useObjectStore, usePermissionStore } from '@/store';
 import { Permissions } from '@/utils/constants';
 import { ButtonMode } from '@/utils/enums';
 import { formatDateLong } from '@/utils/formatters';
+import { objectService } from '@/services';
 
 import type { Ref } from 'vue';
 import type { COMSObject } from '@/types';
@@ -24,6 +25,10 @@ import type { COMSObject } from '@/types';
 type COMSObjectDataSource = {
   lastUpdatedDate?: string;
 } & COMSObject;
+
+type DataTableObjectSource = {
+  [key: string]: any;
+};
 
 // Props
 type Props = {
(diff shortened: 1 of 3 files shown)


frontend/src/components/object/ObjectTable.vue
```diff
@@ -85,18 +96,47 @@ async function showPermissions(objectId: string) {
 
 watch(getObjects, async () => {
```
TimCsaky, line 97: `getObjects` is the object data that's in the store. We won't need to watch it any more.

frontend/src/components/object/ObjectTable.vue
```diff
@@ -85,18 +96,47 @@ async function showPermissions(objectId: string) {
...
   clearSelected();
+  loadLazyData(lazyParams);
```
TimCsaky, line 99: i dont think passing `lazyParams` to this function is doing anything

frontend/src/components/object/ObjectTable.vue
```diff
@@ -85,18 +96,47 @@ async function showPermissions(objectId: string) {
...
-    return x;
-  });
+  lazyParams.value = {
+    first: 0,
+    rows: 10,
+    sortField: null,
+    sortOrder: null,
+    filters: filters.value
+  };
+  loadLazyData(lazyParams);
```
TimCsaky, line 112: same here. no need to pass the lazyParams because they are defined 'globally' in the component's state `lazyParams.value`.

frontend/src/components/object/ObjectTable.vue
```diff
@@ -85,18 +96,47 @@ async function showPermissions(objectId: string) {
...
+    .searchObjects({
+      bucketId: props.bucketId ? [props.bucketId] : undefined,
+      deleteMarker: false,
+      latest: true,
+      permissions: true,
+      page: lazyParams.value.page ? lazyParams.value.page + 1 : 1,
+      limit: lazyParams.value.rows
+    })
+    .then((r: any) => {
+      loading.value = true;
```
TimCsaky, line 127: maybe this should be moved to start of function (line 116). and maybe removed from the onMounted() function

TimCsaky (commented):
i know this is a draft and you have made changes but i wonder if these suggestions make sense.

frontend/src/components/object/ObjectTable.vue
jatindersingh93: I made it optional so we don't need to pass everytime.

frontend/src/components/object/ObjectTable.vue
jatindersingh93: I made it optional param

frontend/src/components/object/ObjectTable.vue
jatindersingh93: It has been readjusted.

PR #175, opened 2024-02-16
Refactor filters and tags to work without objectstore

jatindersingh93:
Replace Refresh Button
Make sure data-table updates when object permissions is updated using permission box
Download works on select one and multiple files
Delete works on select one and multiple files
Upload buttons works
Select all works
Detail box works
Remove any orphan codes
[
SC3518](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3518)

# Description

## Checklist

### DIFF
--- a/frontend/src/components/object/ObjectFilters.vue
+++ b/frontend/src/components/object/ObjectFilters.vue
@@ -30,7 +30,6 @@ const metadataStore = useMetadataStore();
 const objectStore = useObjectStore();
 const tagStore = useTagStore();
 const { getMetadataSearchResults } = storeToRefs(useMetadataStore());
-const { getUnfilteredObjectIds } = storeToRefs(useObjectStore());
 const { getTagSearchResults } = storeToRefs(useTagStore());
 const { getUserId } = storeToRefs(useAuthStore());
 
@@ -58,11 +57,7 @@ objectStore.$onAction(({ name, args }) => {
 
 // Computed
 const metadataValues = computed(() => {
-  // Filter out any tags that don't have an objectID that exist in getUnfilteredObjectIds
-  const filteredVals = getMetadataSearchResults.value.filter((searchRes) =>
-    getUnfilteredObjectIds.value.some((obj) => obj === searchRes.objectId)
-  );
-
+  const filteredVals = getMetadataSearchResults.value;
   return (
     filteredVals
       // Take the metadata for the objects, and flatten them into a single array
@@ -76,10 +71,7 @@ const metadataValues = computed(() => {
 });
 
 const tagsetValues = computed(() => {
-  // Filter out any tags that don't have an objectID that exist in getUnfilteredObjectIds
-  const filteredVals = getTagSearchResults.value.filter((searchRes) =>
-    getUnfilteredObjectIds.value.some((obj) => obj === searchRes.objectId)
-  );
+  const filteredVals = getTagSearchResults.value;
 
   return (
     filteredVals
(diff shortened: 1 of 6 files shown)


frontend/src/components/object/ObjectFilters.vue
```diff
@@ -137,13 +130,13 @@ const selectedFilterValuesChanged = () => {
 
 const searchMetadata = async () => {
   searching.value = true;
-  await metadataStore.searchMetadata();
+  await metadataStore.searchMetadata(props['bucketId']);
```
kyle1morel, line 133: `props.bucketId`

frontend/src/components/object/ObjectFilters.vue
```diff
@@ -137,13 +130,13 @@ const selectedFilterValuesChanged = () => {
...
   searching.value = false;
 };
 
 const searchTagging = async () => {
   searching.value = true;
-  await tagStore.searchTagging();
+  await tagStore.searchTagging(props['bucketId']);
```
kyle1morel, line 139: `props.bucketId`

frontend/src/store/metadataStore.ts
```diff
@@ -87,10 +87,10 @@ export const useMetadataStore = defineStore('metadata', () => {
     }
   }
 
-  async function searchMetadata(metadataSet: Array<MetadataPair> = []) {
+  async function searchMetadata(bucketId: string = '', metadataSet: Array<MetadataPair> = []) {
```
jujaga, line 90: bucketId should be the second argument, and probably should not hard default to empty string when unspecified, as that will limit this store action's utility in the long run in the event you need to do a COMS-wide search.

frontend/src/store/tagStore.ts
```diff
@@ -82,11 +82,11 @@ export const useTagStore = defineStore('tag', () => {
     }
   }
 
-  async function searchTagging(tagset: Array<Tag> = []) {
+  async function searchTagging(bucketId: string = '', tagset: Array<Tag> = []) {
```
jujaga, line 85: bucketId should be the second argument, and probably should not hard default to empty string when unspecified, as that will limit this store action's utility in the long run in the event you need to do a COMS-wide search.

frontend/src/components/object/ObjectTable.vue
```diff
@@ -117,7 +119,6 @@ const loadLazyData = (event?: any) => {
       tableData.value = r.data;
       totalRecords.value = +r?.headers['x-total-rows'];
       loading.value = false;
-
       // add to object store
       r.data.forEach((o: COMSObject) => {
         objectStore.$patch((state) => {
```
kyle1morel, line 124: Not a change from your PR, but we should avoid using globals like this and instead turn this into a store action.

frontend/src/components/object/ObjectTable.vue
```diff
@@ -173,34 +188,53 @@ const selectedFilters = (payload: any) => {
...
+      @select-all-change="
+        (e) => {
+          selectAll = e.checked;
+          if (e.checked) {
+            selectCurrentPage();
+          } else {
+            objectStore.setSelectedObjects([]);
+          }
+        }
+      "
```
kyle1morel, line 237: As Tim mentioned we may be able to remove this logic as we are now always selecting the entire payload.

TimCsaky (approved):
there are some unused imports but otherwise looks good to me.

jujaga (approved):

PR #177, opened 2024-02-24
Refactor File Details page

TimCsaky:
# Description

- reduce coupling with prop/watch pattern
- use models to pass current and latest versionId's
- convert some store 'find' actions into getters
- extend action to fetchMetadata/tags for all versions in one shot
- consolidate async action calls for COMS data

## Types of changes

## Checklist

### DIFF
--- a/frontend/src/components/object/ObjectVersion.vue
+++ b/frontend/src/components/object/ObjectVersion.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import { storeToRefs } from 'pinia';
-import { onMounted, ref, watch } from 'vue';
+import { watch, computed } from 'vue';
 import { useRouter } from 'vue-router';
 
 import { DeleteObjectButton, DownloadObjectButton } from '@/components/object';
@@ -11,7 +11,7 @@ import { ButtonMode } from '@/utils/enums';
 import { formatDateLong } from '@/utils/formatters';
 
 import type { Ref } from 'vue';
-import type { User, Version } from '@/types';
+import type { Version } from '@/types';
 
 type VersionDataSource = {
   createdByName?: string;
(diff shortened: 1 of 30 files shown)


frontend/src/components/object/ObjectVersion.vue
```diff
@@ -21,72 +20,46 @@ type VersionDataSource = {
...
-watch(getVersions, async () => {
-  const versions = versionStore.findVersionsByObjectId(props.objectId);
-  await userStore.fetchUsers({ userId: versions.map((x: Version) => x.createdBy) });
-  tableData.value = versions.map((v: Version, index, arr) => ({
-    ...v,
-    createdByName: getUserSearch.value.find((u: User) => u.userId === v.createdBy)?.fullName,
-    versionNumber: arr.length - index
-  }));
+watch(props, async () => {
+  await userStore.fetchUsers({ userId: versions.value.map((x: Version) => x.createdBy) });
```
jujaga, line 62: Not completely certain if we really need the async and await here as this callback just fires and finishes when it finishes. There doesn't appear to be any dependent functions that need to be blocking on this specific thread.

frontend/src/store/userStore.ts
```diff
@@ -31,7 +31,20 @@ export const useUserStore = defineStore('user', () => {
   // Getters
   const getters = {
     getIdps: computed(() => state.idps.value),
-    getUserSearch: computed(() => state.userSearch.value)
+    // returns a single User for provided user ID
+    getUser: computed(
+      () => (userId: string | undefined) => state.userSearch.value.find((x: User) => userId === x.userId)
+    ),
+    // returns an array of Users corresponding to provided user IDs
+    getUsers: computed(() => (userIds: Array<string> | undefined) => {
```
jujaga, line 39: Not absolutely certain, but assuming no typescript snafus, it might be better to have the blah | undefined be marked as blah? instead to signal optionality?

kyle1morel (approved):

frontend/src/views/detail/DetailObjectsView.vue
jujaga, line 1: I'm presuming this is an intentional change to the file details page where no matter what, we want routing to always reroute a url to go from /details?objectId=blah to /details?objectId=blah&versionId=foo in every case? We want to make sure this url isn't too hard-coerced and can still forward-hydrate as this will be the main entrypoint for numerous user flows.

jujaga (approved):
Some minor flow comments and potential optimizations regarding async/await. Looks reasonably good otherwise - good work! Being able to finally use computeds over the action workaround will definitely simply things in the long run.

jujaga (approved):

norrisng-bc (approved):

jujaga:
Temporarily closing and reopening due to pipeline issues

PR #179, opened 2024-03-05
Implement Magic Link redirection support and InviteButton component

jatindersingh93:
# Description

## Checklist

### DIFF
--- a/frontend/src/components/common/InviteButton.vue
+++ b/frontend/src/components/common/InviteButton.vue
@@ -6,7 +6,7 @@ import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 import ShareLinkContent from '@/components/object/share/ShareLinkContent.vue';
 import { Button, Dialog, TabView, TabPanel, RadioButton, InputText, useToast, InputSwitch } from '@/lib/primevue';
 import { useConfigStore, useObjectStore, useBucketStore } from '@/store';
-import { permissionService } from '@/services';
+import { inviteService } from '@/services';
 
 import type { Ref } from 'vue';
 import type { COMSObject, Bucket } from '@/types';
@@ -74,7 +74,7 @@ async function sendInvite() {
     if (formData.value.expiresAt) {
       expiresAt = Math.floor(Date.now() / 1000) + formData.value.expiresAt;
     }
-    const permissionToken = await permissionService.createInvite(
+    const permissionToken = await inviteService.createInvite(
       props.bucketId,
       formData.value.email,
       expiresAt,
@@ -89,9 +89,9 @@ async function sendInvite() {
 }
 onMounted(() => {
   if (props.objectId) {
-    obj.value = objectStore.findObjectById(props.objectId);
+    obj.value = objectStore.getObject(props.objectId);
   } else if (props.bucketId) {
-    bucket.value = bucketStore.findBucketById(props.bucketId);
+    bucket.value = bucketStore.getBucket(props.bucketId);
   }
 });
 </script>
--- a/frontend/src/components/common/InviteButton.vue
+++ b/frontend/src/components/common/InviteButton.vue
@@ -152,12 +152,12 @@ onMounted(() => {
           >
             <RadioButton
               v-model="formData.expiresAt"
-              :input-id="value"
+              :input-id="value.toString()"
               :name="name"
               :value="value"
             />
             <label
-              :for="value"
+              :for="value.toString()"
               class="ml-2"
             >
               {{ name }}
--- a/frontend/src/components/common/InviteButton.vue
+++ b/frontend/src/components/common/InviteButton.vue
@@ -5,11 +5,11 @@ import { computed, ref, onMounted } from 'vue';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 import ShareLinkContent from '@/components/object/share/ShareLinkContent.vue';
 import { Button, Dialog, TabView, TabPanel, RadioButton, InputText, useToast, InputSwitch } from '@/lib/primevue';
-import { useConfigStore, useObjectStore } from '@/store';
+import { useConfigStore, useObjectStore, useBucketStore } from '@/store';
 import { permissionService } from '@/services';
 
 import type { Ref } from 'vue';
-import type { COMSObject } from '@/types';
+import type { COMSObject, Bucket } from '@/types';
 
 // Props
 type Props = {
@@ -34,11 +34,13 @@ const props = withDefaults(defineProps<Props>(), {
 
 // Store
 const objectStore = useObjectStore();
+const bucketStore = useBucketStore();
 const { getConfig } = storeToRefs(useConfigStore());
 const toast = useToast();
 
 // State
 const obj: Ref<COMSObject | undefined> = ref(undefined);
+const bucket: Ref<Bucket | undefined> = ref(undefined);
 const isRestricted: Ref<boolean> = ref(props.restricted);
 const showInviteLink: Ref<boolean> = ref(false);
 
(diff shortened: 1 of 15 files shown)


frontend/src/components/common/InviteButton.vue
```diff
@@ -0,0 +1,228 @@
...
+const displayInviteDialog = ref(false);
+
+// Share link
+const bcBoxLink = computed(() => {
+  const path = props.objectId ? `detail/objects?objectId=${props.objectId}` : `list/objects?bucketId=${props.bucketId}`;
+  return `${window.location.origin}/${path}`;
+});
+const comsUrl = computed(() => {
+  return `${getConfig.value.coms?.apiPath}/object/${props.objectId}`;
+});
```
jujaga, line 74: This shouldn't be something we need to be caching in the component - this looks like something that should be residing in the Service layer. Not completely sure but on quick glance this invite button shouldn't be needing to send a direct link into coms on Line 136?

frontend/src/components/common/InviteButton.vue
```diff
@@ -0,0 +1,228 @@
...
+
+//Action
+async function sendInvite() {
+  try {
+    let expiresAt;
+    if (formData.value.expiresAt) {
+      expiresAt = Math.floor(Date.now() / 1000) + formData.value.expiresAt;
+    }
+    const permissionToken = await permissionService.createInvite(
```
jujaga, line 83: We should get this to use the inviteService instead of the permissionService.

frontend/src/components/common/InviteButton.vue
```diff
@@ -0,0 +1,228 @@
...
+  </Button>
+</template>
+
+<style scoped lang="scss">
+h2 {
+  font-weight: bold;
+}
+ul {
+  padding-left: 22px;
+}
```
jujaga, line 227: While these style changes suffice to get the job done, it would be better to instead assign a class or self-made construct to modify these styles instead of modifying the base type itself - in the future if this component were to be expanded and reuse these same types, they could cause some undesired effects for styling.

frontend/src/services/permissionService.ts
```diff
@@ -102,5 +102,22 @@ export default {
...
+   * @param {string} expiration timestamp for token
+   * @returns {string} objectId
+   */
+  async createInvite(bucketId?: string, email?: string, expiresAt?: number, objectId?: string) {
+    return comsAxios().post(`${PATH}/invite`, {
+      bucketId: bucketId || undefined,
+      email: email || undefined,
+      expiresAt: expiresAt || undefined,
+      objectId: objectId || undefined
+    });
```
jujaga, line 121: Let's move and replace the createInvite in inviteService with this definition.

frontend/src/types/InviteCreateOptions.ts
```diff
@@ -0,0 +1,6 @@
+export type InviteCreateOptions = {
+  bucketId?: string;
+  email: string;
+  expiresAt?: string;
+  objectId?: string;
+};
```
jujaga, line 6: If we wanted to be absolutely precise about this and represent the XOR relationship between bucketId and objectId, we could pull off something like what is described in this SO: <https://stackoverflow.com/a/53229567>

frontend/src/views/invite/InviteView.vue
```diff
@@ -0,0 +1,133 @@
...
+  403: {
+    // wrong email
+    title: 'Forbidden',
+    detail: 'The person who sent you the link may have restricted it to another email address'
+  },
+  404: {
+    // invitation not found/expired
+    title: 'Not found',
+    detail: 'Invitation not found or has expired'
+  },
```
jujaga, line 58: If the invitation is expired, COMS will report a 410. We might need to define a 410 key with the same contents as 404.

jujaga:
Temporarily closing and reopening due to pipeline issues

frontend/src/views/invite/InviteView.vue
```diff
@@ -0,0 +1,133 @@
...
+          }
+        });
+      }
+    })
+    .catch((error: any) => {
+      const errData = error?.response?.data;
+      if (errData) {
+        toast.error(errData.status, errData.detail);
+        title.value = errorMessage[errData.status]['title'];
+        detail.value = errorMessage[errData.status]['detail'];
```
wilwong89, line 92: If a status code comes up that's not part of the pre-planned error messages, it will cause the "Processing..." message to stay displayed. Perhaps updating the pre-planned messages and using option chaining (eg. `errorMessage?.[errData.status]['title']`) to fix.

jujaga: We're also borderline in the space of potentially using a switch/case operator instead; that way in the event you get another unexpected/uncoded HTTP response, you could fallback to the 'default' case and have it output something terminal instead of hanging on "Processing..."

frontend/src/services/inviteService.ts
```diff
@@ -1,27 +1,33 @@
...
+   * @param {string} bucketId
+   * @param {string} objectId
+   * @param {string} email to be used for access
+   * @param {string} expiration timestamp for token
+   * @returns {string} uuid token of the invite
    */
-  createInvite(params: InviteCreateOptions): Promise<string> {
-    return comsAxios().get(`${PATH}`, {
-      params: params
+  async createInvite(bucketId?: string, email?: string, expiresAt?: number, objectId?: string) {
```
jujaga, line 15: These functions do not need the async keyword on them as an await is not being used anywhere in the body.

frontend/src/services/inviteService.ts
```diff
@@ -1,27 +1,33 @@
...
   },
 
   /**
    * @function getInvite
    * Use an invite token
+   * @param {string} token uuid
    * @returns {Promise} An axios response
    */
-  getInvite(token: string): Promise<any> {
+  async getInvite(token: string): Promise<any> {
```
jujaga, line 30: These functions do not need the async keyword on them as an await is not being used anywhere in the body.

frontend/src/services/inviteService.ts
```diff
@@ -1,27 +1,33 @@
...
    */
-  createInvite(params: InviteCreateOptions): Promise<string> {
-    return comsAxios().get(`${PATH}`, {
-      params: params
+  async createInvite(bucketId?: string, email?: string, expiresAt?: number, objectId?: string) {
+    return comsAxios().post(`${PATH}/invite`, {
+      bucketId: bucketId || undefined,
+      email: email || undefined,
+      expiresAt: expiresAt || undefined,
+      objectId: objectId || undefined
```
jujaga, line 20: In this situation, it might be better to use the ?? operator instead of the || operator.

frontend/src/views/invite/InviteView.vue
```diff
@@ -41,29 +33,31 @@ const detail: Ref<string> = ref('Please check your link again');
...
+      break;
+    case 409:
+      title.value = 'Not found';
+      detail.value = 'Invalid object or bucket';
+      break;
+    case 410:
+      title.value = 'Gone';
+      detail.value = 'Invitation has expired';
+      break;
+    default:
```
jujaga, line 58: We need a default fallback value title and detail error message for when the error code is not one of the above (for some reason or another). I suggest the following:
```ts
default:
  title.value = 'Error';
  detail.value = 'Something has gone horribly wrong :(';
  break;
```

jujaga (approved):

frontend/src/components/common/InviteButton.vue
```diff
@@ -0,0 +1,222 @@
+<script setup lang="ts">
+import { storeToRefs } from 'pinia';
+import { computed, ref, onMounted } from 'vue';
+
+import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
+import ShareLinkContent from '@/components/object/share/ShareLinkContent.vue';
+import { Button, Dialog, TabView, TabPanel, RadioButton, InputText, useToast, InputSwitch } from '@/lib/primevue';
+import { useConfigStore, useObjectStore, useBucketStore } from '@/store';
+import { inviteService } from '@/services';
```
kyle1morel, line 9: Import ordering

frontend/src/components/common/InviteButton.vue
```diff
@@ -0,0 +1,222 @@
...
+import type { COMSObject, Bucket } from '@/types';
+
+// Props
+type Props = {
+  bucketId?: string;
+  objectId?: string;
+  labelText: string;
+  restricted?: boolean;
+};
+export type inviteFormData = {
```
kyle1morel, line 21: Type names should start with an uppercase

frontend/src/components/common/InviteButton.vue
```diff
@@ -0,0 +1,222 @@
...
+    );
+    inviteLink.value = `${window.location.origin}/invite/${permissionToken.data}`;
+    toast.success('', 'Invite link is created.');
+    showInviteLink.value = true;
+  } catch (error: any) {
+    toast.error('', 'Error getting token');
+  }
+}
+onMounted(() => {
+  if (props.objectId) {
```
kyle1morel, line 91: Consider: New lines between functions. Makes it easier to parse the file.

kyle1morel (approved):

PR #198, opened 2024-05-15
Display confirmation dialog and toasts on toggling file to public

norrisng-bc:
# Description

Setting the "set to public" switch to "on" (in both the object table and file permissions modal) now opens an additional modal that asks the user to confirm before doing so:

Screenshot 2024-05-15 at 4 10 24 PM

(this dialog is only displayed when toggling a file from private to public)

A toast is then displayed: 

Screenshot 2024-05-15 at 4 39 19 PM

A toast is also displayed when toggling a file from public to private:

Screenshot 2024-05-15 at 4 38 50 PM

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3667

## Types of changes

## Checklist

## Further comments

N/A

### DIFF
--- a/frontend/src/components/object/ObjectPermission.vue
+++ b/frontend/src/components/object/ObjectPermission.vue
@@ -66,6 +66,7 @@ onBeforeMount(() => {
         class="ml-4"
         :bucket-id="object.bucketId"
         :object-id="object.id"
+        :object-name="object.name"
         :object-public="object.public"
         :user-id="getUserId"
       />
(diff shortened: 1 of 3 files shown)


frontend/src/components/object/ObjectPermission.vue
```diff
@@ -66,6 +66,7 @@ onBeforeMount(() => {
         class="ml-4"
         :bucket-id="object.bucketId"
         :object-id="object.id"
+        :object-name="object.name"
         :object-public="object.public"
```
TimCsaky, line 70: i guess we could just pass the whole `object` in one binding (`:object="object"`) instead of 4 separate ones, and break it down in the child component, and i think we've found some other ways since adopting this pattern early on.. but this is fine for now.

TimCsaky (approved):

PR #199, opened 2024-05-17
Add email form validation

jatindersingh93:
- Add Email validation to the invite link
- Validate using Regex
- Notify if email syntax is not correct
- currently not using required, we may deal in a separate ticket as it may require conditional schema

[SC3599](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3599)

# Description

## Checklist

### DIFF
--- a/frontend/src/components/common/InviteButton.vue
+++ b/frontend/src/components/common/InviteButton.vue
@@ -2,22 +2,16 @@
 import { storeToRefs } from 'pinia';
 import { computed, ref, watch } from 'vue';
 
+import { Form } from 'vee-validate';
+import { object, string } from 'yup';
+import TextInput from '@/components/form/TextInput.vue';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 import ShareLinkContent from '@/components/object/share/ShareLinkContent.vue';
-import {
-  Button,
-  Dialog,
-  TabView,
-  TabPanel,
-  RadioButton,
-  Checkbox,
-  InputText,
-  useToast,
-  InputSwitch
-} from '@/lib/primevue';
+
+import { Button, Dialog, TabView, TabPanel, RadioButton, Checkbox, useToast, InputSwitch } from '@/lib/primevue';
 import { Spinner } from '@/components/layout';
 
-import { Permissions } from '@/utils/constants';
+import { Permissions, Regex } from '@/utils/constants';
 import { inviteService } from '@/services';
 import { useAuthStore, useConfigStore, useObjectStore, usePermissionStore, useBucketStore } from '@/store';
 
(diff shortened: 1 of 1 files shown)


frontend/src/components/common/InviteButton.vue
```diff
@@ -109,6 +103,11 @@ const isOptionUnselectable = (optionName: string) => {
   return optionName === 'Read';
 };
 
+//Validation
+const schema = object({
+  Email: string().matches(new RegExp(Regex.EMAIL), 'Provide a valid email address')
+});
```
TimCsaky, line 109: we should do: 'if restrict toggle is on, email field is required' rule
we have an example of how to implement this in previous code.

frontend/src/components/common/InviteButton.vue
```diff
@@ -2,25 +2,19 @@
...
-  InputSwitch
-} from '@/lib/primevue';
+import { Button, Dialog, TabView, TabPanel, RadioButton, Checkbox, useToast, InputSwitch } from '@/lib/primevue';
 import { Permissions } from '@/utils/constants';
 import { inviteService } from '@/services';
 import { useAuthStore, useConfigStore, useObjectStore, usePermissionStore, useBucketStore } from '@/store';
 
 import type { Ref } from 'vue';
 import type { COMSObject, Bucket } from '@/types';
+import { Regex } from '@/utils/constants';
```
TimCsaky, line 17: the `Regex` member can be imported in line 11

frontend/src/components/common/InviteButton.vue
```diff
@@ -201,85 +199,89 @@ onMounted(() => {
...
-        <div class="p-inputgroup mb-4">
-          <Button
-            class="p-button p-button-primary"
-            @click="sendInvite"
-          >
-            <font-awesome-icon
-              icon="fa fa-envelope"
-              class="mr-2"
+          <p class="mt-4 mb-2">Restrict to user's email</p>
+          <div class="flex flex-wrap gap-5 mb-3">
```
TimCsaky, line 251: when you rebase you'll see the structure of this template has changed a bit

frontend/src/components/common/InviteButton.vue
```diff
@@ -201,85 +199,89 @@ onMounted(() => {
...
+            <InputSwitch
+              v-model="isRestricted"
+              class="mt-1"
+            />
+            <!-- Add email validation -->
+            <TextInput
+              v-if="isRestricted"
+              v-model="formData.email"
+              name="Email"
```
TimCsaky, line 260: i see you changed this. either Email or inviteEmail is fine by me

frontend/src/components/common/InviteButton.vue
TimCsaky: it will be something like this:
```js
const schema = object({
      isRestricted: boolean(),
      Email: string().when("isRestricted", {
        is: true,
        then: string().matches(new RegExp(Regex.EMAIL), 'Provide a valid email address'),
      }),
    });
 ```

PR #202, opened 2024-05-22
Add invite email form validation

jatindersingh93:
- Add Email validation to the invite link
- Validate using Regex
- Notify if email syntax is not correct
- currently not using required, we may deal in a separate ticket as it may require conditional schema

[SC3599](https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3599)

# Description

## Checklist

### DIFF
--- a/frontend/src/components/common/InviteButton.vue
+++ b/frontend/src/components/common/InviteButton.vue
@@ -2,22 +2,15 @@
 import { storeToRefs } from 'pinia';
 import { computed, ref, watch } from 'vue';
 
+import { Form } from 'vee-validate';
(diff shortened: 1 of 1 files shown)


norrisng-bc (commented):
https://github.com/bcgov/bcbox/blob/15ac1094f8f23e0138a981a46a6f33982f9ed47d/frontend/src/components/common/InviteButton.vue#L139

Looks like this line isn't correctly grabbing the email field in the form - it returns `undefined`, so `emails` never gets populated with the entered email address and the call to the BCBox `/email` endpoint is never made.

The invite link gets created just fine though.

PR #211, opened 2024-06-07
Add tab navigation and aria labels for permissions model

jatindersingh93:
- Make folder and file permissions keyboard accessible
- Make permission table keyboard and screen able accessible
- Focus back to the button element when model is closed
- Add aria tags for screen reader
- Make content in model screen reader friendly
- Keep focus in the model so outside DOM elements are not available

# Description

## Checklist


TimCsaky (approved):
Great job. good to have this pattern to re-use

PR #215, opened 2024-06-18
Show results of bulk invite and add/remove in table

norrisng-bc:
# Description

This PR adds a `DataTable` (contained inside a modal) displaying the results of a bulk operation (i.e. add/remove permissions, sending invites). This is displayed after the user clicks "Submit" and all operations are complete.

Each operation is displayed as one of 3 possible result types:
* Success (green checkmark) - operation was successful
* No action (grey minus sign) - no action was performed (e.g. permission already exists on user)
* Failure (red "X") - fail (see "further comments" at bottom)

The actual bulk operations were implemented in #212, though the results weren't displayed in a user-friendly table.

The contents of this `DataTable` can also be downloaded as a CSV file.

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3677
https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3679

## Types of changes

## Checklist

## Further comments

Given how bulk add/remove/invite works at the moment, there will never be errors to display:
* CHES does not return any emailing-related errors (e.g. if the receiving address rejects the email). 
* Improperly-formatted email addresses (i.e. RFC 5322 non-compliant) are already caught by the form validation before the CHES API call is amde.
* The parts of the UI for bulk add/remove/invite permissions are already inaccessible if the user doesn't have the required permissions, so insufficient user permissions should not be an issue.

### DIFF
--- a/frontend/src/components/common/BulkPermissionResults.vue
+++ b/frontend/src/components/common/BulkPermissionResults.vue
@@ -1,12 +1,18 @@
 <script setup lang="ts">
-import { Dialog } from '@/lib/primevue';
+import { ref } from 'vue';
+import { Column, DataTable, Dialog } from '@/lib/primevue';
+import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
+
 // Props
 type Props = {
   results: Array<Object>;
   resource: any;
   resourceType: string;
 };
-const props = withDefaults(defineProps<Props>(), {});
+
+const props = withDefaults(defineProps<Props>(), {
+  results: undefined
+});
 
 const modelValue = defineModel<boolean>({ default: false });
 </script>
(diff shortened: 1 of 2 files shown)


frontend/src/components/common/BulkPermissionResults.vue
```diff
@@ -1,21 +1,101 @@
...
+                v-if="data.status === 1"
+                icon="fa-circle-check"
+                class="icon-success"
+              />
+              <font-awesome-icon
+                v-else-if="data.status === 0"
+                icon="fa-circle-minus"
+                class="icon-noaction"
+              />
+              <!--
```
TimCsaky, line 72: not sue if we need to include this comment. but if you want to leave it in that's fine by me. 
technically CHES can in theory return an error (for example if CHES is down or we arent authorized) that we can show (but it will occur for all invites).

frontend/src/components/common/BulkPermissionResults.vue
```diff
@@ -1,21 +1,101 @@
...
 <template>
   <Dialog
     v-model:visible="modelValue"
     header="Results"
     :modal="true"
   >
-    <pre>{{ props.results }}</pre>
+    <DataTable
+      ref="batchResults"
+      :export-filename="`${resource.name}_bulk_results`"
```
TimCsaky, line 34: change this line to:
```js
:export-filename="`${ resourceType === 'object' ? resource.name.replace(/\.[^/.]+$/, '') : resource.bucketName }_bulk_results`"
```

frontend/src/components/common/BulkPermissionResults.vue
```diff
@@ -1,21 +1,101 @@
...
+const props = withDefaults(defineProps<Props>(), {
+  results: undefined
+});
 
 const modelValue = defineModel<boolean>({ default: false });
+
+// Exports DataTable results
+const batchResults = ref();
+
+const exportCSV = (event: any) => {
```
TimCsaky, line 22: guess we can remove the argument

frontend/src/components/common/BulkPermissionResults.vue
norrisng-bc: True, the "further comments" for this PR are probably sufficient.

frontend/src/components/common/BulkPermissionResults.vue
norrisng-bc: I had to throw in the redundant param here - the tests don't seem to like how the`@click=exportCSV($event)` in the `<Button>` has one argument but the declaration has 0 params.

TimCsaky (approved):

frontend/src/components/common/BulkPermissionResults.vue
norrisng-bc: Fixed in #216.

PR #216, opened 2024-06-18
Bulk results modal: change "save results" icon to match BCBox style

norrisng-bc:
# Description

This tweaks the styling of the "save results" button icon for the bulk results modal, so that it matches the style throughout the rest of BCBox (i.e. blue icon with no background). Previously, it was a white icon on a blue button.

This also removes the unneeded argument in `exportCSV()`, as mentioned in https://github.com/bcgov/bcbox/pull/215#discussion_r1644943689.

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-3679

## Types of changes

## Checklist

## Further comments

In the interest of time, #215 was pushed to production without these fixes. The current PR is a minor continuation and includes them.

### DIFF
--- a/frontend/src/components/common/BulkPermissionResults.vue
+++ b/frontend/src/components/common/BulkPermissionResults.vue
@@ -19,7 +19,7 @@ const modelValue = defineModel<boolean>({ default: false });
 // Exports DataTable results
 const batchResults = ref();
 
-const exportCSV = (event: any) => {
+const exportCSV = () => {
   batchResults.value.exportCSV();
 };
 </script>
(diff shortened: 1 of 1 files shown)


norrisng-bc:
> I would suggest to add background color to the hover over to keep it consistent with close button Here is something already exists
> 
> `.p-dialog .p-dialog-header .p-dialog-header-icon:enabled:hover { color: #495057; border-color: transparent; background: #e9ecef; }`

I based my button off the existing action buttons (i.e. the ones on the right of the object table) - I don't think this kind of shading is present there. I'd probably defer to Tyler on that, to be honest.

PR #225, opened 2024-12-06
allow file upload process to create new version of file if it exiists

TimCsaky:
# Description

allow file upload process to create new version of file if it exists

### DIFF
--- a/frontend/src/components/object/ObjectUpload.vue
+++ b/frontend/src/components/object/ObjectUpload.vue
@@ -55,12 +55,17 @@ const onUpload = async (event: any) => {
 
           const data = formData.find((x: ObjectMetadataTagFormType) => x.filename === file.name);
 
-          await objectStore.createObject(
+          const response = await objectStore.createObject(
             file,
             { metadata: data?.metadata },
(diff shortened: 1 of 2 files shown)


frontend/src/components/object/ObjectUpload.vue
```diff
@@ -55,12 +55,16 @@ const onUpload = async (event: any) => {
...
-          await objectStore.createObject(
+          const response = await objectStore.createObject(
             file,
             { metadata: data?.metadata },
             { bucketId: bucketId, tagset: data?.tagset },
             { timeout: 0 } // Infinite timeout for big files upload to avoid timeout error
           );
+
+          // show toast for any object updates
+          if (response?.newVersionId) toast.info(`New version created for ${file.name}`,'');
```
norrisng-bc, line 66: I'd probably mention something along the lines of "Previously deleted file `{filename}`found. This has been overwritten by the latest upload".

Definitely needs a bit of wordsmithing, but the idea is that if the previously soft-deleted file isn't visible anywhere inside BCBox, it could cause user confusion if we're all of a sudden implying that it already exists.

PR #249, opened 2026-04-16
Fix bug where search IDP permissions chunking code was unreachable; fine-tune urlLimit for COMS API calls

norrisng-bc:
# Description

When searching IDP permissions, the code that decides whether to chunk COMS API calls runs only if there are multiple bucketId's. However, this code was unreachable due to a hardcoded condition that always resolves to False. This has been fixed in this PR.

Additionally, the calculation of the `urlLimit` before chunking is required has now been fine-tuned. It now accounts for the length of every `idp` and `permCode` that is actually present in the URL.

Without this fix, BCBox makes giant API calls with 999 `bucketId[]`'s, which confuses COMS and causes an error where it doesn't send the appropriate CORS headers, which in turn causes BCBox to fail. 

Such massive query arrays were prevoiusly blocked by Joi, but bcgov/common-object-management-service#318 removed a hidden 20-item limit, so implementing that PR had the effect of exposing the error that this PR addresses.

https://apps.nrs.gov.bc.ca/int/jira/browse/SHOWCASE-4279

## Types of changes

## Checklist

## Further comments

N/A

### DIFF
--- a/frontend/src/services/permissionService.ts
+++ b/frontend/src/services/permissionService.ts
@@ -201,13 +201,8 @@ export default {
   async bucketSearchIdpPermissions(params?: any) {
     // Normalize bucketId to an array if it's a string
     if (params.bucketId && typeof params.bucketId === 'string') params.bucketId = [params.bucketId];
-    // if searching with more than one bucketId, but only one idp
-    if (
-      params.bucketId &&
-      params.bucketId.length > 1 &&
-      (Array.isArray(params.idp) ? params.idp.length === 1 : true) &&
-      1 > 2
-    ) {
+    // if searching with more than one bucketId
+    if (params.bucketId && params.bucketId.length > 1) {
       /**
        * split calls to COMS if query params (eg bucketId's)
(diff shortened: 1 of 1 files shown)


frontend/src/services/permissionService.ts
```diff
@@ -218,9 +213,17 @@ export default {
       const baseUrl = new URL(`${new ConfigService().getConfig().coms.apiPath}${PATH}/${IDP}/${BUCKET}`).toString();
       urlLimit -= baseUrl.length;
 
+      // if param is just a string, coerce it into an array, but don't do anything if it's not in URL at all
+      const idpList = params.idp ? (Array.isArray(params.idp) ? params.idp : [params.idp]) : undefined;
```
TimCsaky: might be clearer:
```js
const idpList = typeof params.idp === 'string' ? [params.idp] : params.idp;
const permCodeList = typeof params.permCode=== 'string' ? [params.permCode] : params.permCode;
```

TimCsaky (approved):

